Issue #6, opened 2026-06-03
Speed Benchmarking Script


m-zain-khawaja:
# Description

Please write a script which loads dummy data and performs GPU warm-up followed by model inference speed benchmarking. This script should reside in the Models/speed_benchmark folder.

m-zain-khawaja:
Thanks @Arseni10Lk

PR #7, opened 2026-06-03
feat(benchmarking): add speed benchmarking

Arseni10Lk:
* Added the script that loads dummy data, runs a forward pass, and prints the benchmarking results.
* Described the script in the appropriate README.
* Modified the main README so that other contributors can share their benchmarking results.

Completes #6 
@m-zain-khawaja

### DIFF
--- a/Model/speed_benchmark/README.md
+++ b/Model/speed_benchmark/README.md
@@ -1,3 +1,15 @@
 # Speed Benchmark
 
-A script to load dummy data, warm-up the GPU and perform inference on 100 samples to calculate inference speed benchmarks of the model
+`speed_benchmarking.py` is a script to load dummy data, warm up the GPU, and perform inference on 100 samples to calculate inference speed benchmarks of the model.
+
+## Tracked parameters
+
+The script outputs:
+* Average FPS — Frames per second that the model can process.
+* Average Latency [ms] — The typical delay it takes the GPU to complete a single forward pass of the model from start to finish.
+* Worst-Case Latency [ms] — The 99th percentile latency. It means 99% of your frames were processed faster than this number.
+* Latency Jitter [ms] — This measures the predictability and stability of the inference. By taking your slowest typical frame (p99) and subtracting your fastest typical frame (p1), you get the "spread" or variance in your processing time.
+* Peak VRAM Allocated [MB] — The minimum theoretical footprint your model needs to exist. This is the maximum amount of GPU memory that actually held the forward-pass-related data.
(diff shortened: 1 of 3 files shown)


bharatwrrr:
Thanks for putting this together. A couple of small things worth discussing:

Memory metrics: Since the runs aren't wrapped in `torch.no_grad()`, the reported peak VRAM would be inflated due to the computation graph that PyTorch builds. Since we're benchmarking inference, `no_grad` on the warmup and the timed runs would give us VRAM without the graph, which may be more accurate.

Jitter definition: I researched a bit about this metric, and I found that `p99 - p50` is what people use (apart from standard deviation), largely because it tells you how bad the worst frames are relative to typical behaviour.  `p99 - p1`measures the spread across both tails of the distribution, so a batch of unusually fast runs would inflate the jitter number even if worst-case latency is fine. 

Happy to discuss any of this further.

Arseni10Lk:
Hey @bharatwrrr 

In short I agree with the first one, and the second one is tricky.

1. Since we are benchmarking actual model operation (inference) and not its training, we do not need gradients because we are not changing the weights. So you are completely right, thank you!
2. I had it like that originally for jitter. Then updated it to have it better aligned with what we discussed in the meeting. Now when you put it like that, I think it is definitely more reasonable to use the `p99-p50` approach.

* Counter question would be whether we need latency jitter at all because the worst case jitter is seemingly more important and both the worst case latency and average latency are provided, so I think we can either remove jitter completely at this point or keep it as is. WDYT?

m-zain-khawaja:
@Arseni10Lk and @bharatwrrr  - thank you for the discussion points, my feedback suggestion is:

1. Could we please wrap the inference speed test with torch.no_grad() to have more accurate metrics for the benchmarking.
2. If the common approach for jitter measurement is P99 - P50 then I am happy to go with that too.

Arseni10Lk:
@bharatwrrr @m-zain-khawaja

Done! Additionally, updated the README explanation and added the new results.

Also, added model.eval() because I found that it is the proper way to do it. I did not notice any major difference in performance, but since it is an industry standard, so to speak, I decided to add it.

m-zain-khawaja (approved):
Approved - thanks @Arseni10Lk

m-zain-khawaja:
@Arseni10Lk - we have some model architecture updates with configurable backbones and fusion blocks. I have updated your speed benchmarking script to account for this so that it calculates the benchmarks for each model configuration 👍 

+ @bharatwrrr

Issue #2, opened 2026-06-03
Enhancement Proposal: Multi-Camera Feature Fusion with Batch-Aware Architecture


riita10069:
## Summary

The current `FeatureFusion` module does not perform cross-camera fusion — it only concatenates multi-scale features from a single image along the channel dimension. Additionally, the `DrivingPolicy` module uses `torch.flatten()` without preserving the batch dimension, making mini-batch training impossible. This proposal outlines three approaches to properly unify multi-camera inputs while supporting batched training/inference.

### 1. No cross-camera fusion
`FeatureFusion.forward()` receives multi-scale features with shape `[B*V, C, H, W]` (where V=num_views) and concatenates scales along channels. Each camera's features remain independent — there is no mechanism to reason about spatial relationships across camera views.

### 2. Batch dimension collapse
`DrivingPolicy.forward()` calls `torch.flatten(feature_map)` which collapses all dimensions including batch into a single 1D vector. The Linear layer input sizes (2328) are hardcoded to match exactly `batch=8` (i.e., num_views=8), making it impossible to train with batch_size > 1.

### 3. FutureState outputs per-camera predictions
Since camera features are never unified, `FutureState` predicts future visual features independently per camera rather than for the unified scene.

---

### Approach A: Parametrized View Reshaping (Minimal Change)

Parametrize the model with `num_views` and use reshape/split to handle the view dimension explicitly:

```python
class AutoE2E(nn.Module):
    def __init__(self, num_views=8):
        super().__init__()
        self.num_views = num_views
        ...

    def forward(self, x, visual_history, egomotion_history):
        # x: [B, V, C, H, W]
        B, V, C, H, W = x.shape
        
        # Merge batch and views for backbone processing
        x = x.reshape(B * V, C, H, W)
        features = self.Backbone(x)
        
        # Fuse multi-scale features, then reunify views
        fused = self.FeatureFusion(features)  # [B*V, 1440, 7, 7]
        fused = fused.reshape(B, V * 1440, 7, 7)  # Merge view channels
        
        # Reduce to scene-level representation
        fused = self.view_reduce(fused)  # Conv2d(V*1440, 1440, 1) → [B, 1440, 7, 7]
        ...
```

**Pros:** Minimal architectural change, fixes both the batch dimension bug and enables configurable num_views.  
**Cons:** Simple channel concatenation + Conv does not model spatial relationships between cameras.

---

### Approach B: Cross-Camera Spatial Attention

[...]

PR #12, opened 2026-06-04
feat(architecture):bev spatial cross attention

riita10069:
## Summary

Implement BEV Spatial Cross-Attention fusion mode based on BEVFormer (Li et al., ECCV 2022), selectable as `fusion_mode="bev"` alongside existing `concat` and `cross_attn` modes.

### Key Changes

- **New BEVViewFusion module** (`view_fusion/bev_fusion.py`)
  - Learnable BEV queries (`nn.Embedding`) defining what each BEV grid cell attends to
  - 3D reference points (Z-axis pillar sampling) registered as buffers
  - Portable spatial cross-attention via `F.grid_sample` (no custom CUDA kernels required)
  - Visibility mask to prevent attending to cameras that cannot physically see the corresponding 3D location
  - Pseudo-projection fallback (`nn.Parameter`) enabling training without real camera calibration
- **Propagate `camera_params` through forward pass** — `AutoE2E` → `FeatureFusion` → `ViewFusion` transparently passes intrinsic @ extrinsic matrices for BEV fusion. Existing modes are unaffected.
- **Bug fix**: Feed value-projected features into `grid_sample` instead of raw features (previously `value_proj` received no gradients)
- **Doc fix**: Correct backbone comments from SwinV2 to Swin V1
- **Design Document**: Architecture document covering all 3 fusion modes, design rationale, comparisons, and future directions

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| `F.grid_sample` instead of deformable attention ops | Eliminate CUDA dependency for portability |
| Pseudo-projection (`nn.Parameter`) | Enable end-to-end training before dataset selection |
| Registry pattern for fusion mode switching | Add new strategies without modifying downstream code |
| `camera_params` as Optional kwarg | Maintain backward compatibility with existing code |

## Test Plan

- [x] Existing `concat` / `cross_attn` tests extended to cover `bev` mode — all passing
- [x] BEV-specific tests added:
  - Output shape verification (with and without `camera_params`)
  - Gradient flow to `pseudo_projection` confirmed
  - Gradient flow to `bev_queries` confirmed
  - Different `camera_params` produce different outputs (projection works)
  - `reference_points_3d` shape verification (`bev_h*bev_w, num_z, 3`)
  - No NaN when using pseudo-projection fallback
- [x] `inference_test` exercises BEV mode forward pass with dummy `camera_params`

## References

- [BEVFormer (Li et al., ECCV 2022)](https://arxiv.org/abs/2203.17270)
- [UniAD (Hu et al., CVPR 2023)](https://arxiv.org/abs/2212.10156)
- [PETR (Liu et al., ECCV 2022)](https://arxiv.org/abs/2203.05625)
- Related Issue: #2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__/
 *.py[cod]
+._*
--- a/Design/multi_view_fusion_architecture.md
+++ b/Design/multi_view_fusion_architecture.md
@@ -1,0 +1,546 @@
+# Design Document: Multi-View Fusion Architecture
+
+## Document Metadata
+
+| Field | Value |
+|-------|-------|
+| Status | Proposed |
+| Authors | riita10069 |
+| Created | 2026-06-05 |
+| Related Issue | [#2 - Enhancement Proposal: Multi-Camera Feature Fusion](https://github.com/autowarefoundation/auto_e2e/issues/2) |
+
+---
+
+## 1. Executive Summary
+
+This document describes the design of the multi-view camera fusion system in AutoE2E. The system provides three selectable fusion strategies — Concat, Cross-Camera Attention, and BEV Spatial Cross-Attention — implemented as a pluggable module registry. This phased approach allows researchers to compare fusion strategies on the same codebase and progressively adopt more sophisticated architectures as the project matures.
+
+The design draws heavily from recent advances in vision-centric autonomous driving, particularly BEVFormer [1], UniAD [2], and PETR [3], while maintaining simplicity and portability (no custom CUDA kernels required).
+
+---
+
+## 2. Motivation and Problem Statement
+
+### 2.1 Original Problem
+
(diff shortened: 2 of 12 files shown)


riita10069:
Work in Progress Below Issues

1. `_project_to_2d()` applies `sigmoid()` to the projected 2D coordinates unconditionally. For real camera matrices, projected coordinates should be normalized explicitly using image / feature dimensions, not passed through sigmoid. This destroys out-of-bounds information and makes the visibility mask geometrically invalid.

2. Negative depth is converted to a small positive value via `clamp(min=1e-5)`. Points behind the camera should be masked out, not projected with an artificial positive depth.

3. The visibility mask is computed before adding learned offsets. A sample can be shifted out of bounds by `offset_mean`, but the camera is still counted as visible, while `grid_sample(..., padding_mode="zeros")` silently returns zeros.

4. `num_heads` is effectively collapsed by averaging offsets and weights across heads. This is not a real multi-head deformable/spatial cross-attention implementation.

5. Attention weights are normalized only over points within each camera, then cameras are averaged via `visible_count`. This prevents the module from learning camera-level relevance.

6. The tests are mostly shape/gradient/random-input tests. They do not verify projection correctness, depth masking, out-of-bounds masking, coordinate conventions, or mixed-precision behavior.

m-zain-khawaja (approved):
Approved - thanks @riita10069

Issue #5, opened 2026-06-03
Nvidia Physical AI Dataset Parsing


m-zain-khawaja:
# Description

We wish to parse the Nvidia Physical AI dataset to be able to train our models. Please write a library to ready the data from the dataset and pre-process it to run a forward pass through the network.

riita10069:
I think it's merged

PR #19, opened 2026-06-05
feat (data parsing): add data parsing and forward pass testing for NVIDIA Physical AI dataset

bharatwrrr:
## Data parsing library for NVIDIA PhysicalAI dataset

Addresses #5 

### What's added

- `dataset.py` — `NvidiaAVDataset`, a `torch.utils.data.Dataset` where each item is a valid `(clip_uuid, sample_idx)` pair.
- `camera.py` — decodes camera frames from `.mp4` files using `physical_ai_av.video.SeekVideoReader`. Image preprocessing is derived from the backbone's own config via `timm.data.resolve_model_data_config`.
- `egomotion.py` — parses egomotion using `physical_ai_av.egomotion.EgomotionState`. Downsamples from ~100 Hz to 10 Hz before windowing. Produces both the history tensor `(256,)` and the trajectory supervision target `(128,)` that feeds into the trajectory loss.
- `forward_pass_test.py` — validates the data pipeline with a single forward pass.

### To note:

- **This uses the NVIDIA `physical_ai_av` SDK** (`pip install git+https://github.com/NVlabs/physical_ai_av.git`) for video decoding and egomotion.
- **Egomotion history uses `curvature` directly** (rad/m, available as a column in the egomotion parquet via `EgomotionState`) instead of `yaw_rate` as a proxy. The 4 input signals are `[speed, acceleration, yaw_angle, curvature]`.
- **Trajectory target** is `[acceleration, curvature]` over 64 future timesteps = `(128,)`, matching `DrivingPolicy`'s output.
- **Camera–egomotion alignment** is done by matching timestamps in shared microsecond clip-relative time (egomotion is ~100 Hz, camera is ~30 Hz, neither is perfectly locked).
- **Map tile** (8th view) is a zero placeholder for now. The dataset does not include rendered map tiles. `_make_map_tile` in `camera.py` is the single point to replace when a renderer is available.

@m-zain-khawaja

### DIFF
--- a/Model/data_parsing/README.md
+++ b/Model/data_parsing/README.md
@@ -1,3 +1,22 @@
-# Data Parsing
+# Nvidia Physical AI Data
 
-This folder contains libraries to help parse multiple E2E datsets
+## Data format description
+
+```
+data/
+    camera/
+        camera_front_wide_120fov/
+            <clip_uuid>.camera_front_wide_120fov.mp4
+            <clip_uuid>.camera_front_wide_120fov.timestamps.parquet
+        camera_front_tele_30fov/    ...
+        camera_cross_left_120fov/   ...
+        camera_cross_right_120fov/  ...
+        camera_rear_left_70fov/     ...
+        camera_rear_right_70fov/    ...
+        camera_rear_tele_30fov/     ...
+    labels/
+        egomotion/
+            <clip_uuid>.egomotion.parquet
+```
+
+Each clip is 20s. The egomotion parquet contains 100Hz motion data; camera videos run at 30fps. Both use a shared per-clip timestamp reference (t=0 = clip anchor), which is used to align camera frames to egomotion moments.
--- a/Model/data_parsing/nvidia_physical_ai/README.md
+++ b/Model/data_parsing/nvidia_physical_ai/README.md
@@ -1,1 +1,88 @@
-# Data parsing recipe for [Nvidia Physical AI Dataset](https://huggingface.co/datasets/nvidia/PhysicalAI-Autonomous-Vehicles)
+# Data parsing recipe for [NVIDIA Physical AI Dataset](https://huggingface.co/datasets/nvidia/PhysicalAI-Autonomous-Vehicles)
+
+Parser for the NVIDIA Physical AI dataset, producing tensors directly consumable by AutoE2E's `forward()`.
+
+Download (a subset of) the dataset into `data`.
+
+## Model inputs produced
+
+- `visual_tiles` `(8, 3, H, W)` — 7 camera frames + 1 map tile placeholder
+- `egomotion_history` `(256,)` — 64 past timesteps × 4 signals at 10 Hz
+- `visual_history` `(896,)` — zero-initialised; populated during sequential inference
+- `trajectory_target` `(128,)` — 64 future timesteps × 2 signals (supervision target)
+
+### Egomotion history signals `(256,) = 64 × 4`
+
+- `[0]` Speed (m/s) — `sqrt(vx^2 + vy^2)`
+- `[1]` Acceleration (m/s^2) — `ax`
+- `[2]` Yaw angle (rad) — quaternion → ZYX Euler via `EgomotionState.pose`
+- `[3]` Curvature (rad/m) — `curvature`
+
+### Trajectory target signals `(128,) = 64 × 2`
+
+- `[0]` Acceleration (m/s^2) — `ax`
+- `[1]` Curvature (rad/m) — `curvature`
+
+## Sampling
+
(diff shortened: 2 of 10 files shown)


m-zain-khawaja:
Thanks @bharatwrrr - this looks good to me.

@Arseni10Lk - I believe you also implemented something similar, could you also please take a look regarding the differences between your implementation and this one.

riita10069:
The overall direction looks great to me: the PR adds a useful Dataset wrapper and the tensor shapes broadly match AutoE2E / DrivingPolicy. However, there are several correctness and operability issues:

1. README uses `subset_root`, but `NvidiaAVDataset.__init__` accepts `data_root`. The README usage currently fails.
2. `forward_pass_test.py` instantiates `AutoE2E()`, whose Backbone uses `pretrained=True`; this can fail in offline/CI environments. I would propose to add a pretrained=False path for smoke tests.
3. `__getitem__` repeatedly reads egomotion parquet, camera timestamp parquet, and full mp4 bytes for every sample and camera. This might be prohibitively slow for training. Please cache per-clip egomotion/timestamps and avoid reading full videos per sample.
4. The map tile is documented as zero, but the implementation transforms a black image; after normalization this may not be zero. Use `torch.zeros_like(transform(dummy))` or update the documentation.
5. The trajectory target starts at `sample_idx`, i.e. it includes the current timestep. I would like to confirm whether the target is current+future or strictly future, and adjust indexing/docs accordingly.
6. I think we should add validation for required parquet columns, monotonic timestamps, camera file completeness, and expected number of views.

bharatwrrr:
Thanks, @riita10069. 
Point 3 is a good catch. I'll add functionality to cache egomotion and camera timestamps and rectify the other discrepancies. I found [TorchCodec](https://github.com/meta-pytorch/torchcodec) to be a good candidate for the video reads. But even still, given the size of the dataset, do you think a better option would be to run a preprocessing step to extract frames to JPEG instead to avoid the keyframe decode penalty?

riita10069:
Thanks, that makes sense.

My opinion is that TorchCodec is a good candidate for the immediate fix, especially if we also cache egomotion and camera timestamp metadata and avoid reconstructing readers / re-reading metadata per sample.

For the JPEG extraction idea, I agree that an offline preprocessing path may be better for full-scale training, but I would be slightly cautious about extracting all frames by default.
The reason is that, in this PR’s current sampling design, the model does not actually consume every camera frame. The camera streams are around 30 FPS, while egomotion is downsampled to 10 Hz. A training sample is defined on the 10 Hz egomotion timeline, and only valid `sample_idx` values are used. For a typical 20s clip, that means roughly 200 downsampled egomotion rows and about ~72 valid sample points after reserving 64 timesteps for history and 64 for future target. Meanwhile, the raw camera video would contain around 600 frames per camera at 30 FPS.

So if we extract every camera frame to JPEG, many of those frames may never be used by the current Dataset design. My preference would be to cache only the aligned frames that correspond to valid `(clip_uuid, sample_idx, camera_name)` pairs, and keep a manifest mapping:

`clip_uuid, sample_idx, camera_name, egomotion_timestamp_us, frame_idx, cached_path`

That would avoid paying the repeated keyframe seek/decode cost during training, while also avoiding unnecessary storage blow-up from materializing every 30 FPS frame.

That said, this is my view rather than a blocker for this PR. I think it would be reasonable to merge this PR once the immediate discrepancies are fixed — especially the metadata caching, README/API mismatch, and timestamp handling — and then open a follow-up issue for a proper training-scale frame cache / preprocessing pipeline. 

1. on-the-fly TorchCodec decoding,
2. TorchCodec with cached metadata/frame-index mapping,
3. pre-extracted aligned-frame cache.

I think that would let us keep this PR focused while still tracking the larger data-loading performance question properly.

riita10069 (approved):

bharatwrrr:
@m-zain-khawaja @riita10069, I have rectified most of the discrepancies (except Point 3).

# Change Summary

1. **README `subset_root` vs `data_root`** — fixed parameter name in README to match API.

2. **`pretrained=True` in CI** — added `--no-pretrained` flag to forward pass test, wired `pretrained_backbone` param through `AutoE2E` → `Backbone` → `build_backbone`. I've updated `forward_pass_test` to account for the latest changes in the default backbone and added timers as well.

3. **Per-sample parquet reads** — egomotion dataframes and camera timestamp arrays now cached, passed down to avoid re-reading on every. Video read is left as-is because there are issues with the serializability of TorchCodec objects which may lead to crashes if data loaders with multiple workers are used; follow-up issue to consider sampling images from videos as a preprocessing step to be opened.

4. **Map tile normalization** — replaced `transform(black_image)` with `torch.zeros_like(reference)`.

5. **Trajectory target indexing** — `future_df` now starts at `sample_idx + 1` (strictly future), `max_idx` and `MIN_ROWS` updated accordingly.

6. **Validation** — added `_validate_clip` (camera file completeness), column validation with try/except on parquet read, and `_validate_egomotion_timestamps` (monotonicity check) at dataset construction time.

Thanks.

m-zain-khawaja (approved):
Approved - thanks @bharatwrrr

PR #27, opened 2026-06-06
feat: Added speed benchmark results for NVIDIA RTX 5080

coefinger:
These are the results of running the speed benchmark on a system with an NVIDIA RTX 5080 GPU.
 
The pinned `torch==2.4.1+cu118 (CUDA 11.8)` build predates the Blackwell architecture and ships no `sm_120` kernels, so it cannot run on the RTX 5080. The benchmark was therefore executed against a `CUDA 12.8` build of PyTorch, whose compiled architecture list includes `sm_120`:

  - `torch 2.11.0+cu128`
  - `torchvision 0.26.0+cu128`
  - `CUDA build 12.8`

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,22 @@ To learn more about how to participate in this project, please read the [onboard
 
 </details>
 
+### NVIDIA GeForce RTX 5080 GPU 
+<details open>
+  <summary>Toggle view</summary>
+
+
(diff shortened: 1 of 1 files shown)


m-zain-khawaja (approved):
Approved - thanks @coefinger

PR #28, opened 2026-06-06
perf(tests): reduce test suite time via shared mock backbone fixtures

### LINKED ISSUE #24: Performance: Test Suite Takes ~3 Minutes Due to Redundant Backbone Loading
## Problem

The current test suite (92 tests) takes approximately 177 seconds on a T4 GPU. The speed benchmark (18 configurations) takes over 10 minutes. Both are dominated by backbone forward passes that are not actually testing the backbone itself.

Related: #22 (CI Testing Strategy)

## Analysis

The root cause is that every test instantiates a new `AutoE2E` model, which loads and runs the full backbone (SwinV2 or ConvNeXt) on every forward pass. However, none of the 92 tests are actually verifying backbone correctness — the backbone is pretrained and used as-is. What the tests verify is the behavior of View Fusion, DrivingPolicy, and FutureState.

Rough time breakdown per test:
- Backbone forward (8 views × 256×256): ~80% of time
- View Fusion + DrivingPolicy + FutureState: ~20% of time

With 92 tests each creating a fresh model and running full forward, the backbone is loaded and executed hundreds of times for no testing benefit.

## Ideas to Reduce Test Time

| Approach | Expected Speedup | Change Size |
|----------|-----------------|-------------|
| Session-scoped pytest fixture (reuse model across tests) | ~3× (eliminate repeated model loading) | Small |
| Mock backbone for fusion/policy tests (single Conv that outputs correct shapes) | ~10-20× (bypass backbone forward entirely) | Medium |
| Smaller input resolution for tests (64×64 instead of 256×256) | ~4× (backbone compute scales with resolution) | Small |
| Separate test tiers: unit tests (no backbone) vs integration tests (full forward) | Tier 1 under 10s, Tier 2 stays at ~3min | Medium |

The highest-impact change with least effort would be combining a session-scoped fixture with a mock backbone for non-integration tests. This would bring the majority of tests down to under 10 seconds while keeping a small set of full-forward integration tests for CI validation.

## Current Benchmark Timing (for reference)

The speed benchmark just completed successfully on T4 GPU — full results confirm that a single forward pass at batch=1 takes ~50ms. The test suite runs ~300+ forward passes (92 tests × model fixture parametrized over 3 fusion modes), which accounts for most of the 177 seconds.

## Notes

Not urgent — the tests work fine as-is. But as the test suite grows with new features (temporal fusion, training loop tests, etc.), this will become increasingly painful. Raising it now so we can plan ahead. Would be happy to help implement whichever approach the team prefers.

riita10069:
## Summary

Resolves #24 — Test suite was taking ~3 minutes due to redundant backbone loading on every test.

- Add `conftest.py` with session-scoped mock backbone fixture, eliminating repeated EfficientNet loading
- Refactor tests to use shared fixtures instead of constructing models individually
- Add `pytest.ini` with marker registration for integration test tier
- Separate integration tests (full backbone) from unit tests (mock backbone)

**Result: Test time reduced from ~3 min to ~47 seconds (98 tests passing).**

## Changes

- `Model/tests/conftest.py` — New shared fixtures (`mock_backbone`, `full_model`, `_build_model_with_mock_backbone`)
- `Model/tests/test_auto_e2e.py` — Refactored to use shared fixtures; added `TestFullBackboneIntegration` class
- `Model/pytest.ini` — Marker registration for `integration`
- `Makefile` — Unchanged (original preserved)

## Test plan

- [x] All 98 tests pass on GPU (g4dn.xlarge, T4)
- [x] Test runtime: 47.21s (was ~180s)
- [x] No regressions in existing test coverage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: setup test test-unit test-integration
 
 setup:
-	pip install -r requirements.txt
+	python -m pip install -r requirements.txt
 
 test:
 	cd Model && python -m pytest tests/ -v
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,10 @@
-.PHONY: setup test test-unit test-integration
+.PHONY: setup test benchmark
 
 setup:
-	python -m pip install -r requirements.txt
+	pip install torch timm pytest
 
 test:
-	cd Model && python -m pytest tests/ -v
+	cd Model/tests && python -m pytest test_auto_e2e.py -v
 
-test-unit:
-	cd Model && python -m pytest tests/ -v -m "not integration"
-
-test-integration:
-	cd Model && python -m pytest tests/ -v -m "integration"
+benchmark:
+	cd Model/speed_benchmark && python speed_benchmark.py
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+.PHONY: setup test test-unit test-integration
+
+setup:
+	pip install -r requirements.txt
+
+test:
+	cd Model && python -m pytest tests/ -v
+
+test-unit:
+	cd Model && python -m pytest tests/ -v -m "not integration"
+
+test-integration:
+	cd Model && python -m pytest tests/ -v -m "integration"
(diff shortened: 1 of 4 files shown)


m-zain-khawaja (approved):
Approved - thanks @riita10069

PR #39, opened 2026-06-09
feat(losses): implement trajectory imitation loss

riita10069:
## Summary

Implements #15 — The primary task loss for AutoE2E training.

## Changes

- **Model/model_components/losses/__init__.py** — losses package
- **Model/model_components/losses/trajectory_loss.py** — `TrajectoryImitationLoss` (nn.Module)
- **Model/tests/test_trajectory_loss.py** — Unit tests

## Design

- Supports `smooth_l1` (default) and `mse` loss types
- Optional temporal weighting: exponential decay `λ^t` where `λ` is configurable (default 1.0 = uniform)
- Input: `trajectory_pred (B, 128)`, `trajectory_target (B, 128)` — 64 timesteps × 2 signals [acceleration, curvature]
- Output: scalar loss

```python
from model_components.losses import TrajectoryImitationLoss

criterion = TrajectoryImitationLoss(loss_type="smooth_l1", temporal_decay=0.98)
loss = criterion(trajectory_pred, trajectory_target)
loss.backward()
```

## Test plan

- [x] Unit tests pass (`test_trajectory_loss.py`)
- [x] Output is scalar
- [x] Gradient flows to input
- [x] Temporal weighting changes loss value
- [x] smooth_l1 vs mse produce different values

## Status: Draft

Tests on `main` are broken due to unrelated flexible backbone changes (#36). Will rebase after #36 is merged.

PR #42, opened 2026-06-10
ci: run all test files instead of only test_auto_e2e.py

apgeorg:
The CI workflow pins pytest to a single file:

  ```yaml
  run: cd Model/tests && python -m pytest test_auto_e2e.py -v -m "not integration"
  ```

  As a result, `test_trajectory_loss.py` (added in #39) has never been executed in CI, and any future test file would be silently skipped as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,4 +22,4 @@ jobs:
         run: ruff check
 
       - name: Run unit tests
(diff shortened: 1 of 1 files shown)


riita10069 (approved):

m-zain-khawaja (approved):
Approved - thanks @apgeorg

PR #40, opened 2026-06-10
feat(architecture): replace DrivingPolicy with TrajectoryPlanner and increase BEV resolution to 450x300

riita10069:
## Summary

Resolves #35 — Replaces the flatten-then-MLP `DrivingPolicy` with a `TrajectoryPlanner` that uses ego query + GRU + deformable cross-attention to BEV features, and increases BEV resolution from 7×7 to 450×300 with a forward-biased asymmetric perception range.

## Motivation

The previous `DrivingPolicy` flattened the BEV feature map into a single vector before an MLP. This approach cannot scale beyond small BEV grids — at 450×300×256, flattening would produce 34.5M input dimensions. The literature (UniAD, VAD, BEVFormer) consistently uses ego query + cross-attention for this reason.

The 450×300 grid with 0.4m cell size and forward-biased range (120m front, 60m rear, 60m left/right) was chosen per discussion in #35 to support safety-critical high-speed driving scenarios.

### New: `TrajectoryPlanner` (replaces `DrivingPolicy`)
- Learnable ego query + GRU autoregressive decoder
- At each of 64 timesteps, the GRU hidden state cross-attends to BEV features via deformable attention (`F.grid_sample` — same pattern as `BEVViewFusion`)
- Outputs: `trajectory (B, 128)` + `ego_hidden (B, 256)`
- BEV resolution-agnostic: works with any spatial size

### BEV Resolution
- `BEVViewFusion` default: `bev_h=450, bev_w=300`
- `pc_range=(-60, -60, -5, 120, 60, 3)` — X: -60m to +120m (front-biased), Y: ±60m
- Cell size: 0.4m
- Fully configurable via `view_fusion_kwargs`

### `FutureState`
- Now consumes `ego_hidden (256-dim)` instead of the abolished 14-dim compressed visual vector
- Ego intent injected as per-channel bias over BEV spatial dims

### `AutoE2E.forward()` signature change
```python
# Before:
trajectory, compressed_visual, future = model(x, visual_history, egomotion_history, ...)

# After:
trajectory, ego_hidden, future = model(x, egomotion_history, ...)
```
- `visual_history` input removed
- `compressed_visual_feature_vector` output removed
- `ego_hidden` replaces both

## Files Changed (17 commits)

| Component | Change |
|-----------|--------|
| `trajectory_planner.py` | New module |
| `auto_e2e.py` | DrivingPolicy → TrajectoryPlanner |
[...]

PR #38, opened 2026-06-09
feat(benchmark): standardize result schema and add README generation script

riita10069:
## Summary

Implements #29 — Standardizes benchmark result schema and automates README table generation.

## Changes

- **speed_benchmark.py**: Added `--seed` CLI arg, `driver_version`, `commit_sha`, `input_resolution` to JSON metadata. Results now saved to `results/<gpu-name>.json`.
- **generate_readme_table.py**: New script that reads `results/*.json` and outputs formatted README markdown with environment metadata per GPU section.
- **results/.gitkeep**: Directory for committed benchmark artifacts.
- **README.md**: Updated with new workflow documentation.

## Workflow

```bash
# 1. Run benchmark (saves JSON automatically)
make benchmark

# 2. Commit the JSON artifact
git add Model/speed_benchmark/results/

# 3. Generate README section
python Model/speed_benchmark/generate_readme_table.py
```

## Status: Draft

Tests are broken on `main` due to unrelated flexible backbone changes (#36). Will rebase after #36 is merged.

PR #49, opened 2026-06-10
feat(benchmarking): Add NVIDIA RTX A6000 speed benchmark and update RTX 5080 benchmark results

coefinger:
For the Inference Speed Benchmark, results obtained with an NVIDIA RTX A6000 graphics card have been added, and those for the NVIDIA RTX 5080 have been adapted to the new benchmarking script.


apgeorg:
Hi @coefinger ,

I noticed that you committed two JSON files for the 5080. I think one of them can probably be removed, since there is only a single result shown in the README.

Could you please take a look and confirm whether both files are needed?

coefinger:
Hi @apgeorg,
Thank you for noticing. Yes, there was a duplicate results file, which I removed now.
Sorry for the inconvenience.

apgeorg:
> Hi @apgeorg, Thank you for noticing. Yes, there was a duplicate results file, which I removed now. Sorry for the inconvenience.

No, worries! Dankeschön :)

m-zain-khawaja:
@coefinger - thanks for sharing, and @apgeorg thanks for comments.

I noticed that the metrics are very different for the same hardware RTX 5080. Previously, the model achieved a much higher FPS but now it seems to be much lower.

@riita10069 - do you think there is some discrepancy in the speed benchmarking script?

riita10069:
Thanks for raising this. I checked #38 as well.

My read is that #38 should not explain the large RTX 5080 FPS drop by itself. #38 mostly standardized the benchmark artifact format: seed, driver version, commit SHA, input resolution, JSON result files, and README table generation. It does not materially change the measured forward-pass workload. The benchmark loop still measures the model forward pass with CUDA synchronization around the timed section.

If I compare the old RTX 5080 README numbers with the #49 batch=1 RTX 5080 numbers, the drop is roughly:

* concat / cross-attn: about 27–35% lower
* BEV: about 84% lower

That magnitude looks too large to attribute to #38 alone. My current suspicion is that the bigger factor is the architecture change from #40: DrivingPolicy was replaced with the new TrajectoryPlanner, which runs a 64-step GRU / deformable-attention decoding loop, and BEV resolution was increased to 450x300.

That said, I do think there may still be some unintended behavior either in the benchmark script or in how the #49 results are being reported:

1. avg_fps = 1000 / mean_latency_ms is only correct as samples/sec for batch=1. For batch>1, it should either be renamed to forward-passes/sec or changed to batch_size * 1000 / mean_latency_ms.
2. VRAM reserved may be contaminated by CUDA caching across benchmark cases. Allocated memory is probably more reliable unless each config is run in a fresh subprocess.
3. The #49 numbers were measured at commit 9015914, not directly at the #38 merge commit, so we should not use them to isolate #38’s effect.

So my answer is: I do not think #38 alone explains the regression, but I also do not want to fully rule out a benchmark-script or reporting discrepancy yet.

I may be able to secure an AWS GPU instance. I can try to compare the benchmark script and implementation there, ideally by rerunning the relevant commits under the same cleaned-up benchmark script. That should tell us whether the drop is an expected result of the architecture change, or whether #49 / the benchmark script has some unintended behavior.

m-zain-khawaja:
Thanks for comments @riita10069 - I think you're right, the major difference would be coming in from the new Trajectory Planner module - I didn't consider that. I also expect that the diffusion decoder with flow matching should be considerably faster than the autoregressive decoder we have at the moment 👍

PR #52, opened 2026-06-11
feat(data_parsing): add GPS-to-BEV map tile rendering utility

riita10069:
## Summary

Implements #46 — renders GPS waypoints onto an OpenStreetMap-derived BEV map tile, so datasets without a native map input (KIT Scenes, NVIDIA PhysicalAI) can produce an L2D-style map tile for the model's map view slot.

## Module

```
Model/data_parsing/map_rendering/
├── gps_to_map.py    ← fetch network, map-match, render ego-centric tile, gps_to_tensor
├── cache.py         ← network caching + batch tile pre-rendering
├── test_gps_to_map.py
└── README.md
```

## Key design

**Ego-centric rendering (matches L2D / NVIDIA / KITScenes parsers):** the tile is centered on the ego position with the ego heading pointing straight up. GPS lon/lat is converted to a local metric tangent-plane frame, rotated by `-ego_heading`, and plotted with equal aspect ratio (no lat/lon distortion). A fixed metric window (`±radius_m`) around the ego defines the extent.

**L2D palette:** dark background, gray roads, bright-blue matched route, red raw GPS markers — so a timm backbone transform treats it identically to L2D's shipped map tile.

**Offline preprocessing:** osmnx Overpass fetches are slow and need internet, so this is meant for batch pre-rendering (not in the DataLoader hot path). `cache.py` persists fetched graphs (centroid-quantized for reuse) and renders all tiles for a dataset once.

## Test plan

- [x] 11 tests pass, all with mocked osmnx (no network access required)
- [x] Ego-centric: a point ahead of the ego maps to the upper half; rotating heading rotates the tile
- [x] Map-match fallback (empty route → raw GPS only) verified
- [x] Network cache round-trip verified

## Notes

- Dependencies: osmnx, geopandas, matplotlib (offline preprocessing only)
- Failed clips during batch rendering log a full traceback and are skipped

## Refs

- Closes #46

PR #54, opened 2026-06-11
chore: fix stale test target and add map_rendering test targets

apgeorg:
Fixes the stale `test` target and adds targets for the `map_rendering` tests introduced in #52.

## Changes

```make
test:        # mirrors CI
	python -m pytest Model/tests -v

test-map:    # map_rendering only (needs: pip install matplotlib osmnx)
	cd Model && python -m pytest data_parsing/map_rendering -v

test-local:  # everything runnable on a dev machine
	cd Model && python -m pytest tests data_parsing/map_rendering -v
```

## Verification (local)

- `make test` — 137 passed, 7 deselected (identical selection to CI)
- `make test-map` — 11 passed
- `make test-local` — 148 passed, 7 deselected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
 jobs:
   lint-and-test:
     runs-on: ubuntu-latest
+    env:
+      # `make setup` is platform-neutral; this steers pip to CPU torch wheels.
+      PIP_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


riita10069:
The stale unit-test target is fixed, and keeping map rendering tests outside the default test target is reasonable because they require extra local dependencies. 

Could you consider adding a setup-map or setup-local target so the dependency contract is executable in Makefile.

apgeorg:
> The stale unit-test target is fixed, and keeping map rendering tests outside the default test target is reasonable because they require extra local dependencies.
> 
> Could you consider adding a setup-map or setup-local target so the dependency contract is executable in Makefile.

Agreed! I’ve refined it a bit and added a few cosmetic enhancements as well as some convenience features.

riita10069:
Thanks for sorting this out! I think it'd be good to make `.github/workflows/ci.yml` to use make command.

apgeorg:
> Thanks for sorting this out! I think it'd be good to make `.github/workflows/ci.yml` to use make command.

alright, here we go ;)

riita10069:
Perfect, thank you.

Issue #47, opened 2026-06-10
Proposal: Separate Map Encoder to process Nav-maps and HD-maps


m-zain-khawaja:
## Summary
We can utilize a separate encoder to process map data into BEV features and fuse the map BEV features with image BEV features for downstream trajectory planning. This is more appropriate that treating the map as a type of camera image, since it is already in BEV space. We can have two types of encoders, one which supports rasterized navigation maps, and a second which supports vectorized HD-maps. It is important that we are able to support both Lanelet2 and OpenDrive maps.

<img width="774" height="540" alt="Image" src="https://github.com/user-attachments/assets/5dfe5a64-0ea8-4ad3-8b48-c31548ac2295" />

### Map BEV feature extraction
- VAD map module for vectorized map
- SwinV2 Tiny for navigation map

### Map BEV feature & Image BEV feature fusion
- This can be done with cross-attention between the map/image feature maps
- It is important that we construct the Map BEV space to be the same range dimensions as the Image BEV space to allow for direct alignment between the two BEV grids

riita10069:
Thanks for this proposal. I've been looking into how published models handle this and wanted to share some references that might be relevant to the design decisions.

## Research References

| Pattern | Model | Paper |
|---------|-------|-------|
| Vectorized map queries → BEV cross-attention | VAD | [Jiang et al., ICCV 2023](https://arxiv.org/abs/2303.12077) |
| Separate raster encoder + residual fusion | UMPE | [arXiv 2605.02762, 2025](https://arxiv.org/abs/2605.02762) |
| Polyline discretization for HD maps | lanelet2_ml_converter | [arXiv 2407.17409, 2024](https://arxiv.org/abs/2407.17409) |
| Hierarchical map queries (instance + point) | MapTRv2 | [Liao et al., IJCV 2024](https://arxiv.org/abs/2308.05736) |

## Thoughts on the design

**Raster encoder backbone:**

Your proposal of SwinV2 Tiny makes sense for consistency with the camera backbone. UMPE ([arXiv 2605.02762](https://arxiv.org/abs/2605.02762)) uses ResNet-18 for their rasterized map prior and found it sufficient — but their use case is simpler (single-channel SD map). For richer nav-maps like L2D's waypoint rendering, a more capable backbone like SwinV2 could be beneficial. Perhaps a registry pattern would let us experiment with both and pick whichever works best empirically?

**Fusion method:**

[...]

PR #55, opened 2026-06-12
feat(map_encoder): add raster map encoder and map-image feature fusion

bharatwrrr:
## Separate Map Encoder for Rasterized Nav-Maps
Partially solves #47.

Implements rasterized map encoder for rasterized maps. The map is now supposed to be a separate input with its own encoder branch, fused with image BEV features before the trajectory planner. The data parsers have not been updated in this PR.

---

### What changed

**`model_components/map_encoder/`**:
- `RasterizedMapEncoder`: SwinV2-Tiny; outputs `(B, embed_dim, output_h, output_w)`. Encodes a BEV nav-map image into a spatial feature map; matches feature map shape of BEV features
 
**`model_components/map_encoder/map_bev_fusion/`**:
- `MapCrossAttentionFusion`: image BEV queries attend to map BEV keys/values via pre-norm cross-attention.
- `ResidualMapFusion`: `image_bev + alpha * map_bev` where `alpha` is a per-channel `nn.Parameter` initialized to zero. Map branch contributes nothing at training start; gradient flows once alpha grows (adopted from UMPE (arXiv 2605.02762) based on discussion in #47, but see caveats below).
- Both fusion modes registered in `MAP_FUSION_REGISTRY`.

**`model_components/auto_e2e.py`**:
- `num_views` default changed from 8 to 7.
- `forward()` signature: `(camera_tiles, map_input, visual_history, egomotion_history, ...)`. `map_input` is `(B, 3, 256, 256)`.
- Map encoder output size is derived from the view fusion mode: `(bev_h, bev_w)` for bev mode, `(image_feature_size, image_feature_size)` for concat/cross_attn.

---

**Spatial correspondence caveat.** The residual and cross-attention fusions assume `image_bev[i,j]` and `map_bev[i,j]` correspond to the same location in ego-frame space. This assumption holds cleanly only for `fusion_mode="bev"`, where `BEVViewFusion` builds a metric BEV grid from explicit camera projection matrices and `pc_range`. In that mode both feature maps are anchored to the same ego-frame coordinate system and spatial fusion is geometrically meaningful.

For `fusion_mode="concat"` and `fusion_mode="cross_attn"`, `FeatureFusion` produces an abstract feature map by pooling per-view features; there is no metric spatial correspondence between positions in `image_bev` and positions in `map_bev`. The fusion still works as a learned blending mechanism, but the per-cell spatial alignment that motivates residual and cross-attention fusion in UMPE is not present.

@riita10069 @m-zain-khawaja, Do you think this is actually an issue we should be concerned about?

**Vectorized HD-map encoder deferred.** The registry and `build_map_encoder` interface are already structured to accept a `"vectorized"` key when that encoder is implemented (Lanelet2/OpenDrive, issue #47).

---

### Tests

**`tests/test_map_encoder.py`** - new tests covering:
- `RasterizedMapEncoder`: output shape, non-square outputs, gradient flow, random init verification
- `ResidualMapFusion`: zero-alpha guarantee, per-channel alpha shape, gradient flow
- `MapCrossAttentionFusion`: output shape, map influence on output, gradient flow
- Both registries: unknown key raises `ValueError`, correct types returned
- `TestAutoE2EMapIntegration`: zero-alpha training stability (residual fusion only), map influence with non-zero alpha (residual fusion only), gradient flow through map branch, cross-attn forward pass

### DIFF
--- a/Model/README.md
+++ b/Model/README.md
@@ -1,11 +1,13 @@
 # AutoE2E Architecture
 
 ## Architecture Diagram
+> **Note:** The architecture diagram is outdated and does not reflect the current implementation (separate map encoder branch, BEV fusion as default, residual map fusion). It will be updated in a follow-up PR.
+
 <img src="../Media/auto_e2e_architecture.jpg" width="100%">
 
 ## Inputs and Predictions
 **AutoE2E consumes as input:**
-- 7 camera images at 224x224 resolution (providing a surround view of the vehicle alongside a telephoto front and rear camera for long range viewing)
+- 7 camera images at 256x256 resolution (providing a surround view of the vehicle)
 - Rendered map tile (indicating the high level road network layout and future route of the vehicle)
 - Egomotion history (speed, acceleration, yaw angle and yaw angle rate for the previous 6.4s at 10Hz sampling rate)
 - Visual history (`(896,)` = 64 frames × 14-dim compressed scene memory; provides frame-to-frame visual context, distinct from the planner GRU's intra-trajectory temporal coherence)
@@ -20,7 +22,8 @@
 **Forward signature:**
 ```python
 trajectory, ego_hidden, future_visual_features = model(
-    visual_tiles,        # (B, V, 3, H, W) — 7 cameras + 1 map tile
+    visual_tiles,        # (B, V, 3, H, W) — 7 cameras
+    map_tile,            # (B, 3, H_map, W_map) — BEV nav-map image
     visual_history,      # (B, 896) — frame-to-frame visual memory
     egomotion_history,   # (B, 256)
     camera_params=None,  # (B, V, 3, 4) optional, used by BEV fusion
@@ -29,6 +32,6 @@ trajectory, ego_hidden, future_visual_features = model(
 ```
 
 **To learn the driving policy:**
-- Imitaiton Learning is used to penalize trajectory prediciton as well as World Model Simulation based Reinforcement Learning
+- Imitation Learning is used to penalize trajectory prediction as well as World Model Simulation based Reinforcement Learning
 
 
--- a/Model/model_components/auto_e2e.py
+++ b/Model/model_components/auto_e2e.py
@@ -13,8 +13,12 @@ class AutoE2E(nn.Module):
                  num_timesteps=64, num_signals=2, egomotion_dim=256,
                  visual_history_dim=896,
                  map_type="rasterized", map_in_channels=3,
+<<<<<<< HEAD
                  map_fusion_mode="residual", map_fusion_kwargs=None,
                  planner_mode="gru", planner_kwargs=None):
+=======
+                 map_fusion_mode="residual", map_fusion_kwargs=None):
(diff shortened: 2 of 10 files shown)


m-zain-khawaja:
Thanks @bharatwrrr - I think going forward, BEV fusion makes the most sense and we should adopt it as the default for our architecture, especially in the context of using HD/Navigation maps in the BEV space. It fixes many of the challenges with our earlier architectures namely:

**1. Camera and View agnostic** - the model can flexibly adapt to different camera placements and field-of-view on any vehicle by transferring features to a shared BEV grid

**2. Introspection** - BEV features can easily be decoded into perceptual scene representations - e.g. free-space maps, bounding boxes, lane lines etc and could be used in future downstream modules 

**3. Native map alignment** - the BEV features from the images can be matched 1:1 with the BEV features from the map, helping the network with navigational guidance coupled with situational awareness

riita10069:
Thanks — after thinking through the map-image fusion design, I think the cleanest direction is to make the MapEncoder contract stricter.

Regardless of whether the input is a rasterized nav-map or a future Lanelet2/OpenDRIVE vector map, the MapEncoder should output a metric BEV-aligned feature map:

`map_bev: (B, C, H_bev, W_bev)`

In that framing, I do not think cross-attention needs to be the default fusion mechanism for the rasterized map path. If `image_bev[i, j]` and `map_bev[i, j]` refer to the same physical BEV cell, then a residual or gated additive fusion seems like a simpler and more interpretable baseline:

`fused_bev = image_bev + alpha * map_bev`

or, preferably,

`fused_bev = image_bev + gate * map_bev`

where the gate lets the model decide how much to trust the map prior.

I would still avoid raw unweighted addition, since map priors can be stale, misaligned, partially unavailable, or noisy due to localization and rasterization errors. Recent map-prior fusion work such as UMPE uses alignment-aware and confidence-aware fusion, including zero-initialized residual injection for raster priors and confidence-biased attention / gating for vector priors. UniBEV also supports the idea that weighted BEV-level fusion can be more robust than naive concatenation when modalities are unreliable or missing.

bharatwrrr:
I agree @m-zain-khawaja @riita10069. I've resolved the merge conflicts and made the following changes. Marking this ready for review. Let me know what you think. 

## Changes

- `fusion_mode` default changed from `"concat"` to `"bev"`
- `map_fusion_mode` default changed from `"cross_attn"` to `"residual"`
- BEV fusion falls back to `{"bev_h": 450, "bev_w": 300}` when `view_fusion_kwargs` is not provided
- `map_input` added to all forward calls in `test_auto_e2e.py`
- `MapEncoder.*` parameters exempted from gradient flow assertions in `test_auto_e2e.py` (because zero-init alpha blocks the gradients to the map encoder at training start)
- Gradient flow tests in `test_map_encoder.py` updated to use `mode="train"` with `trajectory_target` after rebase with latest upstream changes
- Session-scoped `map_encoder` fixture added to `test_map_encoder.py` to avoid repeated `timm.create_model` calls
- Updated the signature and image tile shape in the README; added a disclaimer note for the embedded diagram due to lack of edit access.

m-zain-khawaja (approved):
Looks good to me - thanks @bharatwrrr 
Any comments or requests @riita10069?

PR #63, opened 2026-06-16
fix(tests): pass map_input in integration infer-mode calls

apgeorg:
### The bug

PR #55 (merged in `6285a57`) added `map_input` as the 2nd positional argument to `AutoE2E.forward`. The train-mode calls in these tests were updated to pass it, but the two `mode="infer"` calls were missed — so the integration suite has been broken since #55 landed.

Concretely, running the integration suite at `6285a57` fails:

```
$ pytest -m integration
FAILED test_full_forward_pass[concat]    - TypeError: missing 'egomotion_history'
FAILED test_full_forward_pass[cross_attn] - TypeError: missing 'egomotion_history'
FAILED test_full_forward_pass[bev]       - TypeError: missing 'egomotion_history'
FAILED test_resnet50_forward_pass        - TypeError: missing 'egomotion_history'
4 failed, 3 passed
```

The regression went unnoticed because `Model/pytest.ini` deselects the `integration` marker by default (`addopts = -m "not integration"`), so neither the default local run nor CI exercises these tests — the failure only surfaces under an explicit `pytest -m integration`. The 3 passing cases are the `mode="train"` calls, which already pass `map_input` correctly.

### Tests added 
This test closes the gap with a cheap, dependency-free real forward that runs by default. It guards against:

  - **forward-signature regressions** — a reordered/inserted/removed positional arg breaks the real call immediately, in CI rather than only under `-m integration`;
  - **backbone-wiring regressions** — real channel discovery and feature-map shapes (including ResNet50's 5 stages) are verified end-to-end, which the mock backbone cannot do.

  It does not replace the pretrained-weights integration tests (which verify that real weights load); it complements them by making the architecture/forward contract a default-run check.


riita10069:
Only thing to confirm: this PR patches `Model/tests/test_auto_e2e.py`, so if #61 lands first and those tests move to `Model/tests/test_integration.py`, this should be rebased/applied to the split file.

apgeorg:
> Only thing to confirm: this PR patches `Model/tests/test_auto_e2e.py`, so if #61 lands first and those tests move to `Model/tests/test_integration.py`, this should be rebased/applied to the split file.

@riita10069 yes, I will do that

apgeorg:
> Only thing to confirm: this PR patches `Model/tests/test_auto_e2e.py`, so if #61 lands first and those tests move to `Model/tests/test_integration.py`, this should be rebased/applied to the split file.

Alright, @riita10069 the conflicts are resolved

riita10069:
Great! Thanks.

apgeorg:
> Great! Thanks.

Great! So lets merge to have running tests again.. :)

riita10069 (commented):
Merged

PR #65, opened 2026-06-16
Introduce mypy static type-checking

apgeorg:
Adds [mypy](https://mypy-lang.org/) static type-checking to the project and CI.

- New `typecheck` make target (`cd Model && mypy .`), folded into `make ci`.
- Runs in CI between lint and tests.
-  Fixes the errors mypy surfaced

### Config
Lenient by design so typing can be adopted incrementally:
- `ignore_missing_imports` — uninstalled optional deps (osmnx, PIL, …) don't fail the build.
- `follow_imports = silent` — untyped legacy code doesn't block checking.

So mypy only flags **real type errors** today; stricter per-module rules can be added later.


riita10069:
One thing I would adjust before merging: the `camera_timestamps` change seems to weaken an existing invariant. `NvidiaAVDataset` preloads timestamps for all cameras, but `__getitem__` now builds the dict with `.get()`, and `load_camera_frame()` accepts `np.ndarray | None` and falls back to reading the parquet again when a value is missing.

That makes cache misses look like valid fallback behavior, whereas they probably indicate an internal inconsistency. It also risks silently reintroducing per-sample parquet reads.

I’d prefer to keep the stricter invariant:

```python
camera_timestamps = {
    cam_name: self._camera_timestamps[(clip_uuid, cam_name)]
    for cam_name in self.camera_names
}

and keep load_camera_frame(..., camera_timestamps: dict[str, np.ndarray] | None = None) with direct indexing when the cache dict is provided.

apgeorg:
> One thing I would adjust before merging: the `camera_timestamps` change seems to weaken an existing invariant. `NvidiaAVDataset` preloads timestamps for all cameras, but `__getitem__` now builds the dict with `.get()`, and `load_camera_frame()` accepts `np.ndarray | None` and falls back to reading the parquet again when a value is missing.
> 
> That makes cache misses look like valid fallback behavior, whereas they probably indicate an internal inconsistency. It also risks silently reintroducing per-sample parquet reads.
> 
> I’d prefer to keep the stricter invariant:
> 
> ```python
> camera_timestamps = {
>     cam_name: self._camera_timestamps[(clip_uuid, cam_name)]
>     for cam_name in self.camera_names
> }
> 
> and keep load_camera_frame(..., camera_timestamps: dict[str, np.ndarray] | None = None) with direct indexing when the cache dict is provided.
> ```

thank you @riita10069  for your clarification.. reverted!

riita10069 (commented):
Thanks

PR #72, opened 2026-06-17
docs(installation-intruction): added package installation instruction for Windows

Arseni10Lk:
# Problem #

Originally, the instruction steps provided in the README file 
```bash
git clone https://github.com/autowarefoundation/auto_e2e.git
cd auto_e2e
make setup                      # CPU torch wheels
make setup TORCH_CHANNEL=cu118  # or a CUDA build (cu121, ... work too)
```
were not reproducible natively on Windows. 

Moreover, the pip installation covered only the default PyPI torch wheels.

# Fix #

I added a separate instruction for Windows
```bash
git clone https://github.com/autowarefoundation/auto_e2e.git
cd auto_e2e
pip install -r requirements.txt                      # CPU torch wheels
pip install -r requirements.txt --extra-index-url https://download.pytorch.org/whl/cu118  # or a CUDA build (cu121, ... work too)
```

# What is needed # 

I believe that on macOS, it works like on Linux by default. This seems to be confirmed by https://stackoverflow.com/questions/1469994/using-make-on-os-x. But I would want to get a confirmation for this specific setup from somebody who uses macOS.

@m-zain-khawaja

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ To learn more about how to participate in this project, please read the [onboard
 
 Requires **Python 3.12** (the pinned PyTorch build has no wheels for 3.13+).
 
+### Linux ###
+<details open>
+  <summary>Toggle view</summary>
+
 1. **Clone and install dependencies**
 
    ```bash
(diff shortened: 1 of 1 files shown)


apgeorg:
@Arseni10Lk pls check the makefile since our README instructions are based on "make setup" etc. 

According to the makefile, the target commands does not differ from your fix suggestion.

Arseni10Lk:
@apgeorg 

Yes, I understand that. The reason I opened this PR was supporting the installation of required packagew on Windows using plain `pip install ...`. Linux natively supports `make` commands which makes it simpler, but Windows doesn't, so the original (Linux) setup will not be possible on Windows.

So this PR is not really a fix, but rather a clarification for those who use Windows without an extra package manager (which I also outlined).

apgeorg:
Ok, understood. `make` is a tool. It's not available natively on Win and sometimes Linux as well, but it can be installed, just like pip, and both ways work on Windows, Linux, and Mac. Since `make setup` only calls pip under the hood, I suggest adding the plain pip commands as an equivalent path for anyone who doesn't use make. What do you think?

Arseni10Lk:
When you put it like that, it does make a lot of sense. So, agreed, will fix that, thank you for pointing this out!

README.md
```diff
@@ -53,6 +57,24 @@ Requires **Python 3.12** (the pinned PyTorch build has no wheels for 3.13+).
...
+### Windows ###
+<details open>
+  <summary>Toggle view</summary>
+
+**Clone and install dependencies**
+
+```bash
+git clone https://github.com/autowarefoundation/auto_e2e.git
+cd auto_e2e
+pip install -r requirements.txt                      # CPU torch wheels
```
riita10069, line 69: You should use make command like linux.

Arseni10Lk:
@riita10069 @apgeorg Fixed the phrasing. What do you think now?

README.md
apgeorg: @riita10069 the idea here is to provide an alternative way...

riita10069 (approved):

Issue #73, opened 2026-06-17
[Feature]: Trajectory rendering


Arseni10Lk:
## Goal ##
As we discussed during the meeting, it would be beneficial to add trajectory rendering.

It is not urgent; however, it would be helpful as the project progresses.

### Proposed solution

There are several trajectory visualization types that would be useful:

1. Drawing a trajectory on a map tile. This is simple mathematically, but this requires a map tile or at least GPS coordinates so that they can be converted into a map tile.
2. Drawing a trajectory on the camera view. This is more complicated mathematically, but it works on any dataset.

As of now, I am [trying to implement](#74) a map-tile-based solution, but I ran into a [bug/issue](#71).

Ultimately, I think, it would be good to have both implementations.

riita10069:
great work, arseni

Issue #71, opened 2026-06-17
[Bug]: LeRobot version ambiguity/mismatch


Arseni10Lk:
### What happened?

When I run     

```
try:
      dataset = L2DDataset(
          repo_id="yaak-ai/L2D",
          episodes=episodes,
          local_files_only=False,
      )
  except Exception as e:
      print(f"[live] SKIPPED: cannot load dataset: {e}")
      return
```

The error `[live] SKIPPED: cannot load dataset: No module named 'lerobot.common'` shows up. 

This is caused by the fact that LeRobot [does not use "common" in its newest versions](https://huggingface.co/docs/lerobot/lerobot-dataset-v3)

So I think it would help to specify the version for which the scripts were written, currently the dependencies section simply lists `lerobot`
@riita10069 

### Steps to reproduce

```shell
Run `test_live_dataset` function
```

### Traceback / logs

```shell

```

### Device

Both

### Local test status

Haven't run the tests

### Environment

This was recorded for lerobot 5.0.1, 5.0.0 at least

### Confirmation

- [x] I searched existing issues and did not find a duplicate.

riita10069:
I should probably fix this on the repo side, since the dataloader and dependency expectations live there. But if you are interested in addressing it while working on this, I would be happy to review and merge that as well. The important part is to make the supported LeRobot version/path explicit so the behavior is not environment-dependent.

Arseni10Lk:
I would love to try to fix that, but before this, I wanted to understand one thing.

Are the files in the `l2d` folder intended to be run within the same virtual environment as the rest of the files in the repository?

If this is the case, then there is an issue with dependencies and their compatibility:
<img width="426" height="262" alt="Image" src="https://github.com/user-attachments/assets/7f124195-3754-4537-b6aa-3fcc759f4890" />

So something has to be updated, or maybe it can be solved in some other way? I am not sure how to go about that, what do you think?

Arseni10Lk:
So, @riita10069 @m-zain-khawaja, what do you think about that?

m-zain-khawaja:
I suggest we bump up requirements for autoE2E to comply with lerobot 0.5.1 - specifically:
- pythorch >=2.7 , <2.11.0
- torchvision >=0.22.0, <0.26.0
- numpy >=2, <2.4.6

We want to have the same set of requirements for working with the AI model as we do for working with the data

@Arseni10Lk - can you please submit a PR with this fix, thank you.

PR #74, opened 2026-06-17
feat(trajectory-rendering): add trajectory rendering on a map tile

Arseni10Lk:
# Goal #

The aim is to build a function that takes a map tile, trajectory, and velocity as inputs and [builds an image with a predicted trajectory on a map tile](#73)

# Implementation #

1. The velocity and yaw are obtained by integrating acceleration
2. The position is obtained by integrating velocity
3. The position is converted to pixels
4. The trajectory is drawn on a map tile using PIL

# Tests # 

* Currently, the test on dummy data `run_visualization_test_dummy_data` was successfully passed, resulting in the following image:
test_trajectory_output

- [x] Next, the test has to be run based on a LeRobot dataset. Currently, this is halted by [the dataloader issue](#71). The script for this test rests in the `run_visualization_test_l2d` function, but it has not been tested yet.

# Other #
- [x] After the functionality is confirmed, it can be modified to be more visually appealing.
- [x] README will be added to list dependencies and explain the code usage.


riita10069:
Thanks for adding the trajectory overlay utility. The core rendering approach looks reasonable for the ego-centric BEV map convention, but I think this PR should not be merged yet.

Main blockers:
1. The L2D visualization path calls AutoE2E with the wrong signature. AutoE2E.forward requires camera_tiles, map_input, visual_history, and egomotion_history, and inference needs mode="infer". The current call model(visual_tiles, visual_history, egomotion_history) will fail independently of the dataloader issue.
2. The added test file is currently a manual script, not a CI-enforced pytest test. It also swallows exceptions in the dummy path, so failures would not fail the test.
3. The script runs the live L2D path by default and uses local_files_only=False, which may trigger external downloads during a local smoke test.
4. draw.circle requires Pillow >= 10.4.0, but Pillow is not pinned. Please use draw.ellipse for compatibility or add an explicit dependency constraint.

Suggested next step: split this into a pure rendering utility plus a pytest dummy/unit test in Model/tests, then keep the L2D visualization as an optional manual script guarded by --live.

Model/visualization/.gitignore
```diff
@@ -0,0 +1 @@
+*.png
```
riita10069: This local `.gitignore` works if the script is executed from `Model/visualization`, but the output paths in `test_trajectory_render.py` are currently relative to the current working directory. For example, `test_trajectory_output.png`, `test_trajectory_l2d_gt.png`, and `test_trajectory_l2d_pred.png` will be written wherever the script is launched from, not necessarily under `Model/visualization`.

So this ignore rule is a bit fragile: if someone runs the script from the repository root, these PNGs may be generated at the repo root and will not be ignored by `Model/visualization/.gitignore`.

I think we should either:

1. avoid writing generated images into the working tree by converting this into a pytest test that uses `tmp_path`, or
2. make the output directory explicit and then ignore that path from the root `.gitignore`, for example:

```gitignore
# Generated trajectory visualization outputs
/Model/visualization/*.png
```

or, if nested output directories are expected:

```gitignore
/Model/visualization/**/*.png
```

My preference is option 1: use `tmp_path` for tests and avoid needing a new `.gitignore` entry for test artifacts.

Arseni10Lk:
@riita10069 

Thank you for the feedback. I understand all the points except one:

I considered using CI-enforced pytest tests, but this got me thinking. So would a test be just
1. Checking whether the PNG file(-s) with certain properties is(are) generated?
3. Checking whether the image matches a reference image?
4. Both?

Would love to see what you think about this

riita10069:
Thanks for asking. I would not start with a full reference-image comparison.

For this PR, I think the CI-enforced pytest test should mainly verify the behavior of the rendering function, not the exact visual appearance. A good first test would check that:

1. `render_trajectory()` returns a PIL image.
2. The returned image has the same size and mode as the input map image.
3. The input `map_image` is not modified in place.
4. The output image is actually different from the blank input image, meaning something was drawn.
5. For a deterministic dummy trajectory, some expected pixels around the ego/start point or trajectory path are non-background.

Checking only “a PNG file was generated” would be too weak, because the function could produce an empty or incorrect image and still pass. On the other hand, comparing the whole image against a reference PNG can be fragile, because small Pillow/version/rendering differences can cause pixel-level changes even when the behavior is acceptable.

So my preference would be:

* Unit test the rendering function without writing files, using an in-memory dummy map.
* If we want to test file output, use pytest’s `tmp_path` so generated PNGs do not land in the working tree.
* Avoid full golden-image comparison unless we later want a dedicated visual regression test.

Something like this would be enough as a first CI test:

```python
def test_render_trajectory_returns_overlay_without_mutating_input():
    actions = torch.zeros(128)
    actions = actions.view(64, 2)
    actions[:, 0] = 0.5
    actions[:, 1] = 0.01
    actions = actions.flatten()

    map_image = Image.new("RGB", (640, 360), color="#111111")
    before = map_image.copy()

    out = Visualization.render_trajectory(
        action_sequence=actions,
        current_speed=10.0,
        map_image=map_image,
        radius_m=800.0,
    )

    assert isinstance(out, Image.Image)
    assert out.size == map_image.size
    assert out.mode == map_image.mode
    assert list(map_image.getdata()) == list(before.getdata())  # input not mutated
    assert list(out.getdata()) != list(before.getdata())        # something was drawn
```

If we later refactor the metric-to-pixel conversion into a small helper function, we could also test the expected trajectory coordinates directly. That would be even stronger and less brittle than comparing a rendered PNG.

Arseni10Lk:
Here are the current renders

image
image

@m-zain-khawaja 
Now working on the front camera trajectory rendering

Arseni10Lk:
Hello, @m-zain-khawaja 

A problem I am facing right now is that the camera intrinsics do not seem to be provided within the L2D dataset. I can try to just assume something so that it looks reasonably good, but it would be best to have the intrinsics. 

I remember that you said you were in touch with some of the dataset creators (in general, I am not sure about this specific dataset). So, is it possible that they can share the intrinsics if you happen to know that?

m-zain-khawaja:
@Arseni10Lk - I am surprised L2D doesn't provide camera intrinsics, I wasn't aware of that. Are camera extrinsics available? If camera calibration information is altogether missing then we cannot work with this dataset since we require calibration data for BEV transform in the Reactive model and would therefore have to go for the KIT Scenes dataset.

Arseni10Lk:
* Extrinsics are [provided](https://huggingface.co/datasets/yaak-ai/L2D/blob/main/extrinsic_RDF.yaml)
* However, I could not locate the intrinsics; it seems they are not provided
* One more thing is that they do not list the map scale. So, I had to select it through trial and error

So what is the final call? We are then switching to KIT Scenes, right? @m-zain-khawaja

m-zain-khawaja:
@Arseni10Lk - okay, let's go for KITScenes in that case

Arseni10Lk:
With KIT Scenes, the issue I ran into is that KIT Scenes use a [numpy version of less than 2.0](https://github.com/KIT-MRT/kitscenes)

So the question is: what do we do?

The most obvious option is to downgrade our NumPy version. But just wanted to make sure we are aligned on that since this seems to be a major change - @m-zain-khawaja

Arseni10Lk:
Everything is coming together nicely 
image
Just one possible improvement is possible.

@immel-f, is it possible for you to provide the mounting height for the camera? For the image above, I guessed 1.5m, but a more accurate distance would mean more accurate visualization, so if it is not a problem for you, it would be good to have it exactly

riita10069:
Thanks for the updates. After reviewing the current implementation and the existing Flyte/MLflow training pipeline, I think the feature boundary should be changed rather than continuing to extend the dataset-specific live visualization scripts.

I propose moving this functionality into a standalone post-training tool under `Tools/`.

The intended use case is not to instantiate a new model and visualize an arbitrary local Hugging Face sample. It is to inspect the behavior of an actual trained model by consuming the same checkpoint and processed dataset artifacts produced by the training pipeline.

The proposed architecture is:

```text
checkpoint + processed evaluation dataset directory
    ↓
reconstruct the model from checkpoint configuration
    ↓
load model_state_dict
    ↓
run inference in episode and frame order
    ↓
convert predicted and target control sequences into metric trajectories
    ↓
render prediction and ground truth for each frame
    ↓
generate MP4 videos, thumbnails, and a JSON manifest
    ↓
write all artifacts into a self-contained output directory
```

The Flyte workflow integration itself does not need to be included in this PR. However, the tool should be designed so that a later Flyte task can call it directly with downloaded `FlyteFile` and `FlyteDirectory` inputs, return the output as a `FlyteDirectory`, and log the directory through MLflow with something equivalent to:

```python
mlflow.log_artifacts(output_dir, artifact_path="trajectory_visualization")
```

## Proposed scope of this PR

### 1. Standalone tool location

For example:

```text
Tools/trajectory_visualization/
├── cli.py
├── runner.py
├── checkpoint_loader.py
├── dataset_reader.py
├── kinematics.py
├── rendering.py
├── manifest.py
├── README.md
└── tests/
```

The code under `Model/` should remain focused on reusable model and data-processing components. The visualization tool should import those components without modifying the normal training dataset output schema.

### 2. Local filesystem contract

The core tool should accept local paths rather than downloading data directly from Hugging Face or accessing S3 itself.

For example:

```bash
python -m Tools.trajectory_visualization \
    --checkpoint /input/checkpoint/best.pt \
    --dataset-dir /input/evaluation-shards \
    --output-dir /output/trajectory-report \
    --episodes 0 1 2 \
    --max-frames-per-episode 300
```

In Flyte, S3 objects are materialized locally through `FlyteFile.download()` and `FlyteDirectory.download()`. Keeping S3 and Flyte concerns outside the core tool makes it usable both locally and in the platform workflow.

An optional thin S3 adapter could be added later, but it should not be required by the rendering library itself.

### 3. Checkpoint-driven model reconstruction

The tool must not instantiate an unconfigured, randomly initialized model.

The checkpoint should contain or reference all parameters required to reconstruct the model, such as:

```python
{
    "model_state_dict": ...,
    "config": {
        "backbone": ...,
        "num_views": ...,
        "embed_dim": ...,
        "planner_mode": ...,
        "num_timesteps": ...,
        "num_signals": ...,
        "source_hz": ...,
        ...
    }
}
```

The loading path should be approximately:

```python
checkpoint = torch.load(checkpoint_path, map_location=device)
model = AutoE2E(**checkpoint["config"])
model.load_state_dict(checkpoint["model_state_dict"], strict=True)
model.eval()
```

If required configuration is missing or incompatible, the tool should fail explicitly instead of silently using defaults.

### 4. Directory-based sequential dataset reader

The visualization tool should read the processed dataset artifact produced by the platform, rather than constructing `L2DDataset` and downloading data through Hugging Face.

The reader must preserve:

* dataset name and version
* episode identifier
* frame identifier
* timestamp or source frame order
* model input images
* raw or visualization-quality camera/map images
* egomotion history
* trajectory target
* map geometry metadata
* camera calibration when camera projection is enabled

Samples must be grouped by episode and sorted explicitly by frame index or timestamp. The order of tar files, directory entries, or DataLoader workers must not be treated as the temporal contract.

This may require extending the processed dataset manifest or shard metadata. That schema change should be made deliberately, rather than adding visualization-only fields to the default `L2DDataset.__getitem__()` output.

### 5. Separate inference, kinematics, and rendering

The following responsibilities should remain independent:

```text
checkpoint loading
dataset decoding
model inference
control-to-metric conversion
BEV rendering
camera rendering
video/report generation
```

In particular, the renderer should preferably consume metric trajectories:

```python
prediction_xy: Tensor[T, 2]
target_xy: Tensor[T, 2]
```

The conversion from model controls to positions should be a separate function:

```python
controls_to_metric_trajectory(
    controls,
    initial_speed,
    dt,
)
```

This avoids embedding assumptions about `64` timesteps, `2` signals, or `10 Hz` inside the rendering API.

### 6. Artifact output

The output directory should be self-contained and suitable for direct MLflow artifact logging.

For example:

```text
trajectory-report/
├── manifest.json
├── index.html                 # optional static report
├── run-summary.json
└── episodes/
    ├── episode-000042/
    │   ├── video.mp4
    │   ├── thumbnail.jpg
    │   ├── metrics.json
    │   └── frames/
    │       ├── 000000.jpg
    │       └── ...
    └── episode-000108/
        └── ...
```

The manifest should contain enough information for either the MLflow Artifact UI or a future dedicated web application to browse the results:

```json
{
  "schema_version": 1,
  "checkpoint": {
    "name": "best.pt",
    "model_config": {}
  },
  "dataset": {
    "name": "yaak-ai/L2D",
    "version": "..."
  },
  "episodes": [
    {
      "episode_id": 42,
      "start_frame": 100,
      "end_frame": 399,
      "video": "episodes/episode-000042/video.mp4",
      "thumbnail": "episodes/episode-000042/thumbnail.jpg",
      "metrics": "episodes/episode-000042/metrics.json"
    }
  ]
}
```

The MP4 should show prediction and ground truth consistently across frames. At minimum, the legend and color convention must be fixed and documented.

## How this addresses the current review concerns

This architectural change resolves most of the previous blockers:

1. **Visualization-only raw data no longer needs to be added to every normal `L2DDataset` sample.**
2. **The same row is no longer decoded again inside the training dataset solely for visualization.**
3. **The prediction comes from the actual trained checkpoint rather than a randomly initialized model.**
4. **The tool no longer depends on a live Hugging Face download.**
5. **Output paths become explicit and isolated under `--output-dir`.**
6. **Generated files can be logged directly as MLflow artifacts or returned later as a FlyteDirectory.**
7. **The tool can run against exactly the processed dataset and model artifacts used by the platform evaluation pipeline.**
8. **Dataset-specific visualization logic is removed from the core model/data-loading path.**

## Remaining requirements that are not solved only by moving the code

There are still several contracts that must be addressed in this PR.

### A. Map geometry

The tool must not assume that:

* the ego vehicle is at the exact image center,
* the map resolution is `1 m/px`,
* the map axes automatically match the trajectory frame,
* resizing or cropping preserves the original transform.

The dataset artifact must provide an explicit transform, for example:

```python
@dataclass
class MapGeometry:
    meters_per_pixel_x: float
    meters_per_pixel_y: float
    ego_pixel_x: float
    ego_pixel_y: float
    rotation_rad: float
```

If L2D does not provide a verified metric map contract, the map overlay must be marked as approximate or disabled for that dataset.

### B. Camera calibration

A synthetic intrinsic matrix such as `fx = fy = 1000` is not sufficient for a camera trajectory overlay described as perspective-correct.

Camera rendering should require verified calibration data:

* camera intrinsics
* distortion model or rectification status
* extrinsic transform direction
* vehicle/trajectory frame definition
* ground-plane definition

If these values are unavailable, the camera overlay should be omitted or clearly labeled as approximate. The BEV/grid visualization can still be supported independently.

### C. Model output contract

The tool must derive or validate:

* number of timesteps
* number and meaning of output signals
* sampling interval
* acceleration unit
* curvature unit and sign
* speed unit
* coordinate handedness

These values should come from the checkpoint config and dataset manifest. They should not be duplicated as unrelated constants in the renderer.

### D. Evaluation data quality

The currently processed JPEG images may have already passed through normalization, clamping, resizing, or lossy encoding intended for training. A separate visualization representation may be required to produce human-readable output.

The processed artifact should distinguish between:

```text
model_input
visualization_image
```

or document a reversible preprocessing path.

### E. Deterministic scene selection

To compare checkpoints, the visualization tool should support deterministic selection, such as:

* explicit episode IDs
* explicit frame ranges
* a fixed evaluation selection manifest
* worst-ADE/FDE sequences supplied by the evaluation stage

Randomly selecting different scenes for each checkpoint would make visual model comparison unreliable.

### F. Error handling and validation

The CLI should fail clearly for:

* missing checkpoint config
* incompatible state dictionary
* unsupported planner output
* missing episode/frame metadata
* invalid `dt`
* non-positive map resolution
* missing calibration for requested camera rendering
* empty output sequences
* unavailable video codec

Returning `None` and failing later during tuple unpacking or `cv2.imwrite()` should be avoided.

## Suggested acceptance criteria

I would consider the revised implementation ready when:

1. A checkpoint generated by the platform can be loaded with `strict=True`.
2. A local processed dataset directory can be read without Hugging Face access.
3. Samples are inferred in deterministic episode/frame order.
4. Both prediction and target are rendered from the same initial vehicle state.
5. At least one valid MP4, thumbnail, and manifest are generated.
6. The output directory can be logged with `mlflow.log_artifacts()` without additional transformation.
7. Unit tests cover checkpoint reconstruction, ordering, control integration, manifest generation, and non-mutating rendering.
8. Dataset-specific map and camera transforms are either verified or explicitly reported as unsupported/approximate.

With this boundary, this PR provides a useful standalone post-training visualization tool, while a follow-up Platform PR can add the thin Flyte task and MLflow logging integration.

riita10069:
@Arseni10Lk  it's really great work!

immel-f:
> Everything is coming together nicely image Just one possible improvement is possible.
> 
> @immel-f, is it possible for you to provide the mounting height for the camera? For the image above, I guessed 1.5m, but a more accurate distance would mean more accurate visualization, so if it is not a problem for you, it would be good to have it exactly

The lidar is close to 2.1 m above the ground, I think from that you could get the camera height via the calibration. Usually the reference coordinate system for everything in KITScenes is the top lidar though, this is also the coordinate system where the poses are. Are you predicting anything in the vehicle frame? (there is no official vehicle frame extrinsic calibration or poses as mentioned before)

Arseni10Lk:
> Are you predicting anything in the vehicle frame? 

Yes, I think the `ReactiveE2E` model builds its internal world model in the vehicle frame and `bev_fusion.py` is in the ego frame, so the resulting trajectory prediction is also relative to the vehicle frame.

> The lidar is close to 2.1 m above the ground

Thank you

riita10069:
wow, great update, thanks

riita10069:
Could you create Tools/trajectory_visualization/README.md ?
I'm trying to use this Tools/trajectory_visualization, but it doesn't work in my environment.

Arseni10Lk:
I was finally able to render a video.

Today I will try to finish everything up and make a final commit (with updated README obviously)

m-zain-khawaja:
@Arseni10Lk - could you please share a link to the video. Also, can you make a video with just the ground truth data in it as well - I would like to be able to publish some updates and showcase the work we are doing, thanks.

riita10069:
I appreciate for that, and I'm using your branch in my environment
because I'm building DataModelConsole (https://d2itskdqq39tx1.cloudfront.net/) and I want to Integrate your mp4 in the scene viewer.
We have decided to focus only KITScenes Dataset. Please focus only KITScenes and I will Integrate everything and resolve conflict.
so When you finish to fix some of reviews and create README file, please mention me here or our discord.

Arseni10Lk:
@m-zain-khawaja 

Yeah will do that, I was trying to send the video this morning, but was short on time

Arseni10Lk:
@m-zain-khawaja 

Here is the video with both trajectories:
https://github.com/user-attachments/assets/48e35b4f-e0d8-46ad-966d-0f58fa1e8648

And here is just the target trajectory:
https://drive.google.com/file/d/1SW7HOUEGeFZPnCVvDzWZKb3mZDCbNrRt/view?usp=drive_link (it was too big to paste it right here); I increased the length to 60 frames for this one.

How does that look? Should I give you a longer clip, or is there something else needed as well?

Arseni10Lk:
@riita10069 

I believe it is done now

riita10069:
@Arseni10Lk Really great work, I appreciate it, I will resolve conflict and merge them/

riita10069:
@Arseni10Lk @m-zain-khawaja 
The Quolity of model is not good (because it's only 1 epoch model) but I success to integrate #74 and Model Inference and Dashboard.

https://d2itskdqq39tx1.cloudfront.net/datasets/kitscenes?version=v2.2

The PR is #136

riita10069 (approved):

Issue #75, opened 2026-06-18
Discussion: BEV-only trajectory scoring + shifted inference schedule for FlowMatchingPlanner


FLagbusted:
# Proposal: BEV-only trajectory scoring + few-step inference upgrade for `FlowMatchingPlanner`

> Per CONTRIBUTING.md, this is posted as a Discussion before any PR, since
> it touches the trajectory planning module. References PR #40 (driving
> policy / FlowMatchingPlanner), PR #55 (map encoder / MapBEVFusion), issue
> #35 (BEV resolution / hardware target), issue #66 (open-loop eval
> pipeline), and the GoalFlow paper (arXiv:2503.05689) discussed in the
> 17/06 WG meeting.

## Context

`FlowMatchingPlanner` (merged in #40) already does conditional flow
matching with BEV cross-attention and AdaLN modulation, and the map
pipeline (#55) already fuses a rasterized nav-map into that BEV before the
planner runs. What's still missing relative to GoalFlow's design — which
we discussed in detail this week — is everything downstream of trajectory
*generation*: GoalFlow doesn't just sample one trajectory, it samples many,
scores them against a goal-point vocabulary and a learned drivable-area
classifier, and picks the best one.

We were told to get BEV working first, before investing in the full
goal-point vocabulary + learned DAC classifier machinery GoalFlow uses.
This proposal is scoped to exactly that boundary: it adds multi-sample
generation and re-ranking using **only data we already have on disk**
(the rasterized map image, the predicted trajectory's own kinematics) —
zero new training, zero new loss terms, zero dependency on a semantic BEV
head or goal-point clustering that doesn't exist yet.

## What this proposal adds (Phase 1)

**1. `TrajectoryComplianceScorer`** — a wrapper around any `BasePlanner`
that draws `K` samples per scene (vectorized via batch-repeat, one
`forward()` call) and re-ranks them by:

- **Drivable-area compliance**, read directly off the *raw rasterized*
  `map_input` pixels at each waypoint's projected BEV coordinate. This is
  a colour lookup, not a classifier — it stands in for GoalFlow's learned
  DAC score until (if) we build a real semantic BEV head.
- **Kinematic comfort**, penalizing samples whose (acceleration,
  curvature) sequence exceeds configurable comfort bounds.

Selection mirrors GoalFlow's own two modes: `"nearest"` (argmax — pick the
single best-scoring sample) or `"mean"` (softmax-weighted blend across all
K samples).

[...]

Issue #35, opened 2026-06-08
Proposal: Increase BEV Feature Map Resolution


riita10069:
## Context

The current BEV feature map resolution is 7×7, which comes directly from the backbone's final feature map size (256×256 input → 7×7 after 5 stages of downsampling). This was never intentionally designed as the BEV grid resolution — it's just the default backbone output that gets passed through.

At 7×7, if we assume a perception range of ~50m around the vehicle, each cell covers roughly 7m × 7m. A single passenger car (4.5m long) fits entirely within one cell, making it impossible to distinguish spatial relationships between nearby objects.

## How other models handle this

Most published BEV perception models define an explicit BEV grid independent of the backbone output resolution, then use attention or projection to populate it:

| Model | BEV Resolution | Perception Range | Cell Size | Reference |
|-------|---------------|-----------------|-----------|-----------|
| BEVFormer | 200×200 | 51.2m × 51.2m | 0.256m | [Li et al., ECCV 2022](https://arxiv.org/abs/2203.17270) |
| UniAD | 200×200 | 51.2m × 51.2m | 0.256m | [Hu et al., CVPR 2023](https://arxiv.org/abs/2212.10156) |
| BEVDet | 128×128 | 51.2m × 51.2m | 0.4m | [Huang et al., 2022](https://arxiv.org/abs/2112.11790) |
| BEVDepth | 128×128 | 51.2m × 51.2m | 0.4m | [Li et al., AAAI 2023](https://arxiv.org/abs/2206.10092) |
| PETR v2 | 100×100 | 51.2m × 51.2m | 0.512m | [Liu et al., ICCV 2023](https://arxiv.org/abs/2206.01256) |
| VAD | 200×200 | 60m × 30m | 0.3m / 0.15m | [Jiang et al., ICCV 2023](https://arxiv.org/abs/2303.12077) |
| SparseBEV | 200×200 | 51.2m × 51.2m | 0.256m | [Liu et al., ICCV 2023](https://arxiv.org/abs/2308.09244) |
| StreamPETR | 100×100 | 51.2m × 51.2m | 0.512m | [Wang et al., ICCV 2023](https://arxiv.org/abs/2303.11926) |

The common range is **100×100 to 200×200**, giving cell sizes of **0.25m to 0.5m**. This is fine enough to distinguish individual vehicles, lane markings, and pedestrians.

## Proposed Approach

Define an explicit BEV grid in the BEV fusion module, decoupled from backbone output:

```python
[...]

PR #76, opened 2026-06-18
feat(trajectory_planning): add BEV-only multi-sample trajectory scorer and shifted inference schedule

### LINKED ISSUE #148: [Feature]: Map rendering for KITScenes
### Description

Currently, the Lanelet2 HD Map render does not encode the route the vehicle drove. This means that our model doesn't learn how to handle a turn or junction and it cannot decide which route to take. This leads to a high error on drives with a junction. However, the model performs great on straight roads without a junction - meaning that it can learn to drive well. It simply needs prior information to guide the action of the planner to a specific navigational goal.

### Example 1 - Navigating a junction with incorrect prediction

Image

### Example 2 - Navigating a road with very accurate prediction

Image

### Proposed solution

- Adding route information to the Lanelet2 HD map is not trivial
- We can instead try using the GPS->BEV map rendering script [here](https://github.com/autowarefoundation/auto_e2e/tree/main/Model/data_parsing/map_rendering) to render a 2D map and encode the route direction based on the ground truth GPS traces snapped to the map
- We can train the first version of the model on 2D map and do HD map training later

### Alternatives considered

_No response_

FLagbusted:
## Summary

Phase 1, training-free upgrade to the flow-matching driving policy: multi-sample
scoring against the BEV/map data we already have. Implements #75.

## What's in here

- TrajectoryComplianceScorer: zero-trainable-params, draws K samples from any
  BasePlanner, re-ranks by drivable-area compliance + kinematic comfort.
- FlowMatchingPlanner.sample_k_trajectories(): batch-repeat K independent noise
  draws per scene in one forward call.

## Round 1 — riita10069's review, contract bug found

- sample_and_score unpacked planner.forward() as (trajectory, ego_hidden) —
  BasePlanner.forward() only ever returns one tensor. Never caught because the
  test's FakePlanner matched the same wrong contract, and the whole test file
  lived at repo-root tests/ — never collected by CI at all (make test only
  globs Model/tests/). Fixed both the contract and a second latent bug (map
  fixture off-by-one) this uncovered; moved the file into Model/tests/.
- Swapped decode_trajectory_to_xy for the canonical integrate_trajectory, now
  using each sample's real starting speed from egomotion_history instead of a
  flat 5 m/s guess for every scene.
- Removed PROPOSAL_diffusion_driving_policy.md (belongs in Discussion #75).

## Round 2 — #161 landed, calibration redesigned on real geometry

#161 (route-conditioned navigation inputs) resolved #148/#149, but changed
more than the calibration: map_input (rendered RGB) doesn't exist any more —
map_context is a 14-channel semantic raster now. Separately, riita10069's
comment on #17 confirmed BEV segmentation labels are headed to a separate
PandaSet track, not landing on KITScenes route-training soon — so this
couldn't wait on that either. Rebuilt directly on what #161 already ships:

- drivable_area_compliance reads map_context's DRIVABLE_AREA channel directly
  (Model.navigation.geometry.MapChannel, a real binary mask) instead of an
  RGB-tolerance colour heuristic.
- project_xy_to_bev_pixel reimplements NavigationRasterGeometry.ego_to_pixel's
  exact formula for batched torch use — cross-checked against the real
  geometry object in tests, not just internal consistency.
- ScorerConfig's calibration now comes from
  Model.navigation.geometry.DEFAULT_NAVIGATION_GEOMETRY
  (geometry_id="kitscenes-v3-bev-1m-v1", 256×256 @ 1.0 m/px) — Correct me If I am Wrong.
- Renamed map_input → map_context throughout, matching ReactiveE2E's naming.

## Testing done

test_trajectory_scorer.py — 22 tests, including a direct cross-check against
DEFAULT_NAVIGATION_GEOMETRY.ego_to_pixel and a test proving compliance reads
the binary channel, not anything colour-adjacent. test_trajectory_planning.py
still 38 passed/1 GPU-only skip. Rebased cleanly onto main post-#161 (only
overlap was flow_matching_planner.py, non-conflicting). ruff clean.

## Checklist

- [x] Conventional Commit PR title
- [x] References #75
- [x] Tests pass (CPU only)
- [x] ruff check passes
- [x] Rebased onto current main (post-#161)
- [x] Calibration sourced from real code (resolves the #148/#149 blocker)


FLagbusted:
When #17 (BEV seg aux loss) lands, drivable_area_compliance upgrades from pixel-colour lookup to label-based with no API change.

FLagbusted:
@m-zain-khawaja ready for review — here's a quick summary of what's in the branch:

- `TrajectoryComplianceScorer` in `trajectory_scorer.py` — zero-param K-sample re-ranking using BEV pixel-colour drivable-area compliance + kinematic comfort bounds
- `sample_k_trajectories()` added to `FlowMatchingPlanner` — returns `[B, K, 128]` via `repeat_interleave` + single `forward()` call; scoped to `FlowMatchingPlanner` only (GRU can't guarantee independence across K). 12 tests in `TestSampleKTrajectories`
- 5 mypy errors in `waymo_e2e_dataset.py` fixed to unblock `make typecheck`

Two items flagged in the checklist that aren't blocking code review:
- `project_xy_to_bev_pixel` calibration defaults (issue #35 geometry) not yet verified against KITScenes renderer — would appreciate a pointer to whoever owns that
- ADE validation against a real checkpoint deferred pending GPU access

This closes the WG action item from 01/07 (*"submit a version of planner with multiple paths or trajectories, related to #17"*). When #17 lands, 

`drivable_area_compliance` can upgrade from pixel-colour to label-based lookup with no API change.

riita10069:
First of all, thank you for this contribution — and for the way it was made. Grounding the design directly in GoalFlow (arXiv:2503.05689), scoping Phase 1 to what needs zero new training, and honestly flagging your own unverified assumptions in #75 is exactly the kind of rigor we hope to see in this project. We have strong interest in this line of SOTA-informed validation, and both `TrajectoryComplianceScorer` and `sample_k_trajectories` are cleanly designed pieces of work.

With that respect stated, here is where we stand as maintainers.

1. The production model does not use `FlowMatchingPlanner` yet. `AutoE2E` defaults to `planner_mode="bezier"`, and all checkpoints trained through our Flyte pipeline so far are Bezier-based. More importantly, `build_planner` currently emits a RuntimeWarning for `flow_matching`: the training loop regresses an Euler-from-noise rollout with SmoothL1, which is not the flow-matching velocity-MSE objective — `compute_planner_loss` is not wired into `train_il` yet. So a meaningful ADE validation of this scorer is blocked on that work, and on a Bezier vs FlowMatching comparison that justifies switching planners. We would prefer to keep the model simple until that comparison exists. The good news is that main now has the yardstick for exactly that comparison: a checkpoint-only retrospective evaluator (`Model/evaluation/kitscenes_benchmark.py`, runnable via the `wf_evaluate_kitscenes_benchmark` Flyte workflow) that reports 3s/5s ADE/FDE for any checkpoint. The path we would like to converge on is: wire in the proper objective, train a FlowMatching checkpoint, then compare Bezier vs FlowMatching vs FlowMatching+scorer on the KITScenes Benchmark. We would be glad to work through that sequence with you — this PR is a genuinely valuable building block for it.

2. The three items you already flagged yourself remain blockers, but they are now resolvable in-repo: the map renderer lives in `Model/data_parsing/kit_scenes/map.py` (to verify `pixels_per_meter` / `ego_row` / `ego_col` and the `drivable_rgb` palette against the actual rendering convention), and a canonical control-to-XY integration taking the real initial speed exists in `Model/evaluation/metrics.py` (`integrate_trajectory`). After a rebase, we would ask that `decode_trajectory_to_xy` be replaced by that shared utility.

3. Two small housekeeping notes: the PR body mentions the shifted timestep schedule but the current diff only contains `sample_k_trajectories` (one of the two seems stale); tests should live under `Model/tests/`, and we would prefer the proposal text to remain in #75 rather than adding a document at the repository root.

To be clear about the conclusion: this is not a rejection of the idea — it is a request to converge on the KITScenes Benchmark as the shared yardstick before merging. Thank you again for raising the bar here; we look forward to iterating on this together.

FLagbusted:
@riita10069 — rebased, fixed the contract bug as you instructed, did the integrate_trajectory swap with real per-scene speed, and flagged the #148/#149 calibration risk directly in the code. 

One open item before this is done: title/body still promise a "shifted inference schedule" that isn't in the diff. Want me to build it for real (α=1.0 preserves current behavior exactly), or trim the title to match what's here?

Two more things:

This branch and #124 both touch flow_matching_planner.py (this one adds sample_k_trajectories, #124 touches compute_planner_loss + the restored ReasoningCoupling/initial_noise). Should be a normal, visible conflict when they land, not a silent one — just a heads-up. #124 is further along and is the dependency for the benchmark comparison anyway, so should we merge if so maybe recommend that one merges first; I'll rebase this branch onto main afterward.
The FlowMatching+scorer row riita10069 asked for on #124 depends on this PR being in good shape — will follow up here once there's a trained checkpoint to actually score with it and will shamelessly take your offer for computation.

riita10069:

I restored your commits in replacement PR #170 from the sanitized history, with the original commit authorship preserved:
https://github.com/autowarefoundation/auto_e2e/pull/170

PR #80, opened 2026-06-22
feat(world-model): FutureState feature-reconstruction (JEPA) loss + optional action conditioning Feat/future state jepa

gcordova10:
## What

Implements the loss side of **#13** — FutureState feature-reconstruction (JEPA) — as **Priority #1** per the maintainer feedback in #56 (@riita10069).

**Two additive changes, both optional/non-breaking:**

1. `losses/feature_reconstruction_loss.py` — `FeatureReconstructionLoss`: JEPA-style MSE between FutureState's 4 predicted future feature maps and frozen-backbone targets at +1.6/3.2/4.8/6.4 s, with optional temporal weighting and a zero-sum-weights guard.
2. `future_state.py` — optional **action conditioning** (`action_dim=None` ⇒ byte-identical to current). Enables counterfactual "what-if" rollouts by projecting the trajectory into a per-channel bias.

`AutoE2E.forward()` contract unchanged. 16 tests added, 157 total green.

## Files changed
- `Model/model_components/losses/feature_reconstruction_loss.py` (new)
- `Model/model_components/losses/__init__.py` (export)
- `Model/model_components/future_state.py` (optional path, default unchanged)
- `Model/tests/test_feature_reconstruction_loss.py` (16 tests)

Relates to #13. See tracking in #56.

Issue #16, opened 2026-06-05
TODO: Dataset Selection and DataLoader Implementation


riita10069:
## Overview

The model currently runs on random dummy tensors. To move toward actual training, a dataset needs to be chosen and a DataLoader implemented. This issue is to track the discussion and eventual implementation.

## Items to consider

- [ ] Evaluate candidate datasets
- [ ] Consider: camera count, calibration availability, trajectory labels, license compatibility
- [ ] Implement dataset class with appropriate preprocessing
- [ ] Sequential frame access (needed for feature reconstruction loss in #13)
- [ ] Camera calibration extraction (needed for BEV fusion mode)
- [ ] Ego trajectory extraction (needed for trajectory loss in #15)

## Candidates (for discussion)

| Dataset | Cameras | Calibration | Trajectory GT | License |
|---------|---------|-------------|---------------|---------|
| nuScenes | 6 | ✓ | ✓ | CC BY-NC-SA 4.0 |
| Waymo Open | 5 | ✓ | ✓ | Waymo Dataset License |
| KITTI Scenes | 2 | ✓ | Partial | CC BY-NC-SA 3.0 |

## Notes

I noticed the data_parsing/ directory already has placeholder folders for kit_scenes, le_robot_l2d, and nvidia_physical_ai, so it seems like there is already some thinking on this. Would be great to hear from the team which direction is preferred — happy to help with the DataLoader implementation once the dataset is decided.

riita10069:
https://github.com/autowarefoundation/auto_e2e/pull/19 is merged.

m-zain-khawaja:
In terms of datasets, my preference is:

1. [LeRobot L2D Dataset](https://huggingface.co/blog/lerobot-goes-to-driving-school) for our early tests: it seems easy and approachable with multi-view, ground truth trajectories, language, and navigation map render + it is fully Apache 2.0 licence based
2. [KIT Scenes](https://kitscenes.com/) for first benchmarking: they already integrate Autoware and have done benchmarks with UniAD - providing multi-sensor + HD map + language and long-tail
3. I quite like WOD, and I think we can also leverage NuScenes for benchmarking too since many models use it.
4. Not so sure about Argoverse - is it used commonly in E2E benchmarks?

gcordova10:
[...]

Issue #13, opened 2026-06-05
Proposal: Feature Reconstruction Loss for FutureState (Auxiliary Self-Supervised Task)


riita10069:
## Summary

The `FutureState` module currently outputs predicted future visual features (`[B, 1440, 7, 7] × 4`) during the forward pass, but these predictions are never used — there is no loss function that compares them against actual future observations. This issue proposes implementing the Feature Reconstruction Loss to activate FutureState as an auxiliary training signal and to enable introspection of model uncertainty.

## Current State

- `FutureState.forward()` predicts features at t+1.6s, t+3.2s, t+4.8s, t+6.4s
- `FrozenBackbone` exists in the codebase but is unused
- The predicted future features are returned from `AutoE2E.forward()` but discarded

### How it works

During training with sequential frame data, we compare what the model predicted the future would look like against what the future actually looked like:

```
Frame t (current):
  image_t → Backbone → Fusion → FutureState → predicted_features[t+1.6s]

Frame t+16 (actual 1.6s later, at 10Hz):
  image_{t+16} → FrozenBackbone → Fusion → target_features[t+1.6s]

feature_reconstruction_loss = MSE(predicted_features, target_features.detach())
```

### Why predict in feature space, not pixel space

Predicting future images (pixels) forces the model to reconstruct irrelevant details like exact texture, shadow positions, and compression artifacts. JEPA (Joint Embedding Predictive Architecture) [1] proposes predicting in a learned latent space instead, where only semantically meaningful changes are captured. MILE [2] demonstrated this principle specifically for autonomous driving — their latent world model outperformed pixel-prediction baselines.

### Why use a FrozenBackbone for targets

If both the predictor and the target encoder are trainable, the network can trivially minimize loss by collapsing all representations to a constant (representation collapse). Using a frozen or slowly-updating target encoder prevents this. This asymmetric design is established in BYOL [3] and JEPA [1].

Two update strategies for the frozen encoder:
1. Keep at pretrained initialization (simplest)
2. Exponential Moving Average from trainable Backbone:
   ```python
   for p_frozen, p_train in zip(frozen.parameters(), trainable.parameters()):
       p_frozen.data = τ * p_frozen.data + (1 - τ) * p_train.data  # τ ≈ 0.996
   ```

### How this helps DrivingPolicy (auxiliary task)

[...]

Issue #84, opened 2026-06-23
Proposal: Motion-grounded residual world model for FutureState (ego-warp + dynamic residual + visual-history conditioning) Relates to: #13 (FutureState JEPA loss), builds on #80 (loss + action conditioning), #82 (temporal memory).


intisar1020:
## Summary

This proposes a small, additive refinement to how `FutureState` predicts the future BEV, building on the JEPA feature-reconstruction loss from #13 / #80 and the temporal-memory interface from #82. The idea has two parts:

1. **Motion-grounded residual:**  predict the future BEV as `Warp_ego(z_t) + Δ_dynamic`, where a *parameter-free* geometric warp removes the predictable ego-induced motion and `FutureState` learns only the dynamic residual.
2. **Visual-history conditioning:** condition the dynamic residual on the temporal context from #82, so it can estimate agent velocities that a single frame might nt be able to reveal.

This is a sketch for discussion. I am fairly new to the codebase and may be missing context, so corrections are very welcome would be really helpful. 

## Motivation

With the JEPA loss merged in #80, `FutureState` now has a  training signal (predicted future BEV vs. frozen-backbone targets on the actual future frames). Two observations about the current prediction setup:

1. **`FutureState` conditions on a single current frame** (`fused_features` + `ego_hidden`, + optional action from #80): it has no access to *past* visual context.
2. Between `t` and `t+Δ`, the BEV changes for **two very different reasons**, currently modelled as one learned delta:
   - **ego-induced apparent motion:** the whole BEV shifts/rotates because *we* moved. Large, dominant, and *exactly known* from the ego trajectory. This is really a spatial transform, which an additive delta might model poorly.
   - **dynamic motion:**  other agents actually moving. The genuinely uncertain part.

This connects to the "predict deltas" direction I floated earlier (TDV).

## Part 1  Motion-grounded residual

Decompose the future BEV into a known geometric part and a learned dynamic part:

```
z_{t+Δ} ≈ Warp_ego(z_t)                         + Δ_dynamic
          └─ known ego pose change ─┘   └─ FutureState predicts only this ─┘
```

- **`Warp_ego(z_t)`:**  resample the current BEV into the future ego frame using the *known* ego displacement over each horizon. That displacement is the **future** ego pose change over `[t, t+h]`: at **training** it comes from the **ground-truth future egomotion** (the same future window that builds `trajectory_target`)
[...]

PR #85, opened 2026-06-23
feat(world-model): slow World Model branch (JEPA) + AutoE2E wiring (#13)

gcordova10:
## Slow World Model branch (JEPA), wired into `AutoE2E` (opt-in)

> _Updated 2026-06-30: this description was rewritten to reflect the post-refactor re-scope (it previously described the pre-refactor BEV version). See the thread for the full history._

Fills the `World_Action_Model_E2E` placeholder left by the architecture refactor (the `Reactive_E2E` + slow World Model split) with the slow @1 Hz branch agreed on 24/06. **Opt-in and additive**: `enable_world_model=False` by default → `Reactive_E2E` behaviour is byte-identical.

### What it adds
- **`WorldActionModel`** (`world_action_model.py`): per-frame multi-view embedding (`FrameEncoder`, default 224-d) → rolling history → **future multi-view feature prediction** with a **frozen-copy target encoder** (`JepaTargetEncoder(mode="frozen")`, stop-gradient, anti-collapse) and an **L1 feature-reconstruction loss** (reuses `FeatureReconstructionLoss`, #80).
- **History aggregator:** `"concat"` (equal-weight FIFO, default) or opt-in **`"attention"`** (a learnable temporal attention-pool, recency-aware, reused from the closed #87/T3) — same `visual_history_dim` (896) either way.
- **Wiring in `auto_e2e.py`:** the World Model produces the `visual_history` fed to the reactive planner.

### Mapping to the 24/06 decisions
| Interface | Decision (24/06) | Here |
|---|---|---|
| Backbone / target | shared backbone, **frozen** target copy (not EMA) | `FrameEncoder` + `JepaTargetEncoder(mode="frozen")` |
| Horizons | `N_past = N_future = 4` @ 1 Hz | `history_len=4`, `num_future_steps=4` |
| `visual_history` | rolling FIFO → planner | `RollingHistoryBuffer` (len 4 → 896) |
| Loss | feature-recon, **L1**, equal weight | `WorldActionModel.jepa_loss` (`FeatureReconstructionLoss("l1")`); the `train_il` wiring itself is **#13**, not this PR |

### Scope (what is and isn't here)
- **Here:** the World Model module + the `AutoE2E` wiring (opt-in) + tests.
- **Not here (own PRs):** the **JEPA loss in `train_il`** → **#13** (`compute_step_loss`, opens after this); the **1 Hz multi-view dataloader** (past+future frames) → **#95** (#16). This PR does **not** modify `train.py`.

### On the rolling buffer (per review)
The `AutoE2E` rolling buffer is **inference/rollout memory only**: pushed embeddings are **detached** (no autograd graph across steps), and it is per-sequence state (`reset_visual_history()` between sequences). Batched **JEPA training is stateless** and goes through the windowed, differentiable `WorldActionModel` path (`encode_history → aggregate_history → predict_future → jepa_loss`) from #13 — not this buffer. The planner and the future prediction read the same post-push history (aligned context).

### Tests
`test_world_action_model.py` + the building-block tests (`test_jepa_target_encoder.py`, `test_feature_reconstruction_loss.py`); full suite green (230 passed), ruff + mypy clean. Rebased on current `main` (`637ee16`).


m-zain-khawaja:
Hi @gcordova10 - I have done some refactoring to the code. If you please take a look at the new auto_e2e.py file, you will notice that we have a Reactive_E2E model which consumes egomotion history, multi-camera images and rendered map image to output a trajectory.

I have added a placeholder for a World Action Model which is the 'slow thinking' component. This model should consume only multi-view images sampled at a lower frequency (e.g. 1Hz or tunable) - we don't need egomotion_history for this, and the model should encode the past multi-view images into a compressed feature vector called 'visual_history' which is then fed into the Reactive_E2E model. The World Action Model should also predict the future multi-view image features as encoded by the same backbone as used in the World Action Model for the same number of future samples as we have past samples and have a feature reconstruction loss - not on BEV, but on multi-view features themselves.

Here is the new architecture diagram which describes this:

auto_e2e_architecture

gcordova10:
Thanks @m-zain-khawaja, this is a great refactor. The explicit split into `Reactive_E2E` (fast, 10 Hz, BEV) and the `World_Action_Model_E2E` placeholder (slow, 1 Hz) is exactly the slow/fast separation we'd been converging on, and putting the feature-reconstruction loss on **multi-view features** rather than BEV is a cleaner choice — it's closer to V-JEPA's image-feature prediction and keeps BEV as the reactive model's concern.

So I'll **re-scope this PR onto your refactor**. The BEV-side JEPA in #85 was built before the refactor and now conflicts, so rather than salvage that diff I'll rebase onto `f484156` and **implement the `World_Action_Model_E2E` placeholder**, reusing the pieces I already built and tested:
- the **frozen/EMA `JepaTargetEncoder`** (stop-gradient target encoder),
- the **`FeatureReconstructionLoss`** (with the L1 option, V-JEPA-style),
- the **`FeatureCompressor`** for the compressed `visual_history` vector.

The BEV `FutureState` predictor drops out of this PR — it belongs to the reactive side.

Concretely the World Action Model would: take the 1 Hz multi-view stream → its own image backbone → encode the past views into the compressed `visual_history` (fed into `Reactive_E2E`) → predict the next *N* future multi-view features → reconstruction loss against the **same backbone (frozen/EMA)** applied to the actual future views.

A few interface questions so I match your diagram exactly:
1. **Backbone** — does the World Action Model get its **own** image backbone (separate from the reactive one), and the target is a frozen/EMA copy of *that* backbone? (frozen vs EMA — the module already supports both.)
2. **Window / N** — how many past frames at 1 Hz, and is `N_future == N_past` exactly (e.g. 4↔4)? Tunable default?
3. **Feature level** — reconstruction on the **per-view** backbone feature maps `[B, V, C, h, w]` (which backbone stage?), per-view or aggregated?
4. **`visual_history` shape** — the compressed vector fed into `Reactive_E2E`: `[B, visual_history_dim]` (896), pooled across views/time?
5. **Integration** — still into `train_il` (#78) with equal loss weighting? The future multi-view frames will need exporting in `data_processing` (same open item as #13).

Do you prefer I rebase #85 into the World Action Model implementation (and repurpose this PR), or open a fresh PR on top of the refactor? Either way I'll reuse the already-tested target-encoder + reconstruction-loss so we're not starting from zero.

On sequencing: I'll rebase the sibling temporal-memory PR (#87) onto the refactor right away since it's independent. For this World Action Model PR, I'd rather lock the 5 interface points above with you first so I'm not building on assumptions, then push the implementation reusing the already-tested target-encoder + reconstruction-loss + compressor.

> Note for @intisar1020 (#84): the ego-warp residual is a BEV operation, so it now naturally sits on the reactive/BEV side rather than this multi-view World Action Model — let's keep that thread separate.

gcordova10:
@m-zain-khawaja — I've **implemented the `World_Action_Model_E2E` placeholder and wired it into `AutoE2E`**, following your answers in the 24/06 meeting and the commented wiring you left in `auto_e2e.py`. I force-pushed this branch with the full implementation (it now supersedes the old BEV-side diff). Here's exactly what changed and the **one design point I'd like to confirm with you**.

### What's in the PR now
A new `world_action_model.py` (`WorldActionModel`, `FrameEncoder`, `RollingHistoryBuffer`) plus the reusable JEPA pieces I'd already built (`JepaTargetEncoder`, `FeatureCompressor`, L1 option in `FeatureReconstructionLoss`), and the wiring in `auto_e2e.py`. It's **opt-in** (`enable_world_model`, default off → `Reactive_E2E` behaviour is byte-identical), reuses the reactive backbone (one shared backbone; the JEPA target is a frozen copy of it), and the JEPA loss is kept out of the model (computed in the training loop via `WorldActionModel.jepa_loss`).

### Mapping to your 24/06 answers
| Interface (my Q#) | Your answer (24/06) | Implemented |
|---|---|---|
| Backbone | one shared backbone, **online trainable + frozen target copy** | `FrameEncoder` reuses the reactive backbone; `JepaTargetEncoder(mode="frozen")` is a frozen copy (stop-gradient, anti-collapse) |
| Window / N | **N_past = N_future = 4 @ 1 Hz** | `history_len=4`, `num_future_steps=4` |
| `visual_history` | rolling FIFO buffer → fed to `Reactive_E2E` | `RollingHistoryBuffer` (len 4), left-padded, FIFO order |
| Loss | feature-reconstruction, **L1**, equal weighting, in `train_il` | `FeatureReconstructionLoss(loss_type="l1")`, `weight=1.0` |

### Wiring (your `auto_e2e.py`, realized)
```python
visual_embedding, future_state_pred = self.World_Action_Model_E2E(camera_tiles, visual_history)
self.visual_history_buffer.push(visual_embedding)   # circular buffer, size N=4
visual_history = self.visual_history_buffer.visual_history()
trajectory = self.Reactive_E2E(camera_tiles, map_input, visual_history, egomotion_history, ...)
return trajectory                       # inference
return trajectory, future_state_pred    # training
```

### One point I'd like to confirm — future-prediction granularity
Here's where the docs diverge and I want your call before I lock it. The **miro diagram** (24/06) shows a per-frame **224-d embedding** and an Encoded Visual History of 896 (4×224), so I currently predict the future as **pooled 224-d embeddings**. But your earlier comment on this PR ("reconstruction on the **multi-view features themselves**, not BEV") and **`TRIAL.md`** ("Future Visual Features Prediction: `torch.Size([8, 1440, 7, 7])` × 4") describe the JEPA target as the **full backbone feature maps** `[B, 1440, 7, 7]`, not a pooled vector.

So: should the future-feature reconstruction be on
- **(a) feature maps** `[B, 1440(=backbone_channels), 7, 7]` × 4 (per `TRIAL.md` / your comment), or
- **(b) pooled per-frame embeddings** 224 (per the miro)?

`FeatureReconstructionLoss` and `FeatureCompressor` already support feature-map targets, so switching to (a) is a small, contained change — I just want to match whichever you intend. (Relatedly, `TRIAL.md` lists the compressed `visual_history` item as `[14]`, vs the miro's 896 — happy to align that too.)

### Coordination with #93 (@intisar1020)
@intisar1020 proposed the same World Action Model in #93 — flagging here so we pool effort rather than duplicate; happy to co-author (their temporal-attention history head + the backbone-freezing choice are clean drop-ins behind this interface).

### Status
World-model + building-block tests pass locally; mypy/ruff clean. **CI is red only because of #88** (the test infra from the refactor), not this code.

Remaining after the feature-level decision is **not part of this PR**: the left-most block of the architecture diagram — the `Data Loader (MP4 Video at 30 FPS)` that subsamples into the `Video sampled at 1 Hz` (World Model + Reasoning) and `Video sampled at 10 Hz` (Reactive) streams. That's **#16** (*Dataset Selection and DataLoader Implementation*); in particular it needs to export the **future 1 Hz multi-view frames** so the JEPA target encoder has something to compare against for end-to-end training.

gcordova10:
Update — pushed `a6bc2b1`: added an **opt-in temporal attention-pool** as the World Model history aggregator (reusing the temporal self-attention from #87, which I'm closing as superseded — it now lives here).

What's new:
- `HistoryAttentionPool` + `WorldActionModel(history_aggregator=...)`:
  - **`"concat"` (default)** → the equal-weight rolling FIFO, **byte-identical to before**;
  - **`"attention"` (opt-in)** → a learnable, recency-aware pool: a single query attends over the N frame embeddings + temporal positional embeddings, projected back to the same `visual_history_dim` (896) — a drop-in for the planner interface.
- Wired through `AutoE2E` via `world_model_kwargs={"history_aggregator": "attention"}`; the default path is unchanged.
- 21 tests (pool shape/grad, concat = identity, attention transforms, AutoE2E end-to-end with the pool); mypy/ruff clean.

The attention-pool reuses the **BEV-queue temporal self-attention I already had in #87**; it converges with @intisar1020's attention-pool point in #93, so rather than duplicate I've folded mine into the World Model here and will close #87.

@m-zain-khawaja — it's **strictly opt-in** (default `concat-FIFO`, unchanged). The 24/06 meeting only talked the rolling FIFO buffer, so this learnable, recency-aware pool is an **extra I'm proposing** — would you be OK keeping it as an opt-in here? CI here is still gated by #88 (fixed in #94).

gcordova10:
Rebased onto current `main` (now includes the #94 CI fix) — **CI is green**.

This PR is the consolidated **slow World Model branch (JEPA)**, wired into `AutoE2E` as opt-in (`enable_world_model=False` by default, so the reactive-only path is byte-identical):

- `WorldActionModel` realizes the 24/06 design: per-frame multi-view embedding (224) → rolling FIFO buffer (N=4 → 896 *Encoded Visual History*) fed to the reactive planner → future-feature prediction with a **frozen-copy target encoder** + stop-gradient, L1 loss.
- Follows the `auto_e2e.py` wiring sketched here (`visual_embedding, future_state_pred = WorldActionModel(...); buffer.push(...)`); forward returns `trajectory` (infer) / `(trajectory, future_state_pred)` (train with WM on).
- Includes the opt-in temporal **attention-pool** history aggregator (`history_aggregator="attention"`) — the reusable block from the closed #87/T3, behind the same 896 interface. Coordinated with @intisar1020 in #93 (he owns the per-view `FrameEncoder`; this is the history side).

Tests: full suite green (221 passed) + dedicated `test_world_action_model.py`; ruff + mypy clean.

To keep this focused, the two consumers are split into their own PRs:
- **#95** — the 1 Hz sequential multi-view dataloader (#16) that produces the history/future frames.
- **#13** — the JEPA loss in `train_il` (`compute_step_loss`), which I'll open right after this merges (it builds on the `WorldActionModel` API here).

Ready for review.

riita10069:
Thanks for the work here.

The main issue is that the PR description claims train.py / JEPA training integration, but the actual training path does not appear to be wired. `build_model()` never passes `enable_world_model=True`, there is no `--enable-jepa` / JEPA loss weight / future-frame batch path, and the current `model(...)` call in `train.py` still does not match the `AutoE2E.forward(camera_tiles, map_input, visual_history, egomotion_history, ...)` signature because `map_input` is missing.

I am also concerned about the internal `visual_history_buffer` state in `AutoE2E`. In training, this can mix history across unrelated shuffled batches/sequences, and because pushed embeddings are not detached, it may also carry autograd graphs across steps. I think this should either be stateless for training, keyed/reset explicitly per sequence, or at minimum detach/reset very deliberately.

There is also a context-alignment issue: `future_state_pred` is predicted from `vh_prev`, while the reactive planner receives the history after pushing the current embedding. If the JEPA future prediction is meant to be conditioned on the same current history used by the planner, this should be reordered.

Finally, the PR body says “Frozen/EMA target encoder” and “EMA update per step,” but `WorldActionModel` currently hardcodes `mode="frozen"` and there is no EMA update in the training loop. `FeatureCompressor` is also added but does not seem to be connected to the World Action Model path.

gcordova10:
Thanks @riita10069, this is a really helpful review — you're right on all of it, and the root cause was on me: the **PR description was stale** (left over from the pre-refactor BEV version), so it advertised things this branch no longer does. I've rewritten the description and pushed fixes (`870d47d`, rebased on current `main`). Point by point:

1. **train.py / JEPA integration** — correct, this PR does **not** wire training, and the old body wrongly claimed it. Scope: this PR is the `WorldActionModel` module + the opt-in `AutoE2E` wiring only. The JEPA loss in `train_il` (`compute_step_loss`, `--enable-jepa`, future-frame batch path, the `map_input` call fix) is **#13**, which I'll open right after this lands; the 1 Hz dataloader is **#95**. Description updated to say this clearly.

2. **`visual_history_buffer` state / no detach** — fixed. Pushed embeddings are now **detached**, so the buffer never carries an autograd graph across steps. It's documented as **inference/rollout memory only**, per-sequence (`reset_visual_history()` between sequences). Batched **training is stateless**: it uses the windowed, differentiable `WorldActionModel` path (`encode_history → aggregate_history → predict_future → jepa_loss`), not this buffer.

3. **Context alignment (`vh_prev` vs post-push)** — fixed. The planner and the future prediction now read the **same post-push history** (including the current frame), so there's no prev/post-push asymmetry.

4. **Frozen vs EMA** — the code is correct (`mode="frozen"`, matching the 24/06 decision — frozen, not EMA); the **old body** was wrong to mention "EMA update per step". Removed from the description. No EMA in the loop by design.

5. **`FeatureCompressor` not connected** — correct, it was unused here. **Removed** it (and its test) from this PR.

Net: code unchanged in behaviour for the reactive-only default (opt-in, off by default), buffer made training-safe, dead code dropped, description now matches the code. Suite 230 passed, ruff + mypy clean. Thanks again for catching the description drift.

m-zain-khawaja (changes requested):
Thank you @gcordova10 - this is great progress. One suggestion was that the future_state predictor should please predict the feature map from the backbone rather than a vector, so we should basically be reconstructing the future backbone features, e.g. [batch, backbone channels, 8 , 8]. Once this change is made I think we can merge it.

gcordova10:
Done, thanks @m-zain-khawaja — the future prediction now reconstructs the **backbone feature maps** instead of a pooled vector (`0aafc18`):

- `BackboneFeatureMap`: a frame → the backbone's last feature map (views mean-pooled), adaptive-avg-pooled to a fixed **`[B, C, 8, 8]`** grid.
- JEPA target = a **frozen copy** of that extractor (stop-gradient) → the future feature maps to match.
- `FutureFeatureMapPredictor`: `visual_history → N × [B, C, 8, 8]` via a light seed-linear + per-step 1×1 conv heads (avoids a huge `history → C·8·8` linear).
- The L1 `FeatureReconstructionLoss` is shape-agnostic, so it now scores feature maps directly.

The planner's Encoded Visual History (per-frame 224 → 896) is unchanged — only the JEPA prediction/target moved to feature-map space. Suite 230 passed, ruff + mypy clean.

Two small things, both happy to do as **follow-ups** (not to block this merge):
- **Per-view targets?** Right now the target is view-aggregated `[B, C, 8, 8]`; I can switch to per-view `[B, V, C, 8, 8]` if you'd rather reconstruct each camera.
- **Prediction context.** The target is now spatial, but the predictor's input (the 896 Encoded Visual History) is pooled, so it predicts the `8×8` map from a spatially-collapsed context. To get the full benefit of spatial targets, a natural next step is to keep spatial structure in the history/context too (spatial tokens instead of a pooled vector). Glad to take that in a separate PR.

m-zain-khawaja (approved):
Approved - thanks @gcordova10

Issue #70, opened 2026-06-17
[Feature]: Replace static architecture images with Mermaid diagrams to keep them in sync with the code


apgeorg:
### Problem / motivation

The architecture is currently documented with static raster images (`Media/auto_e2e_architecture.jpg`) embedded in both `README.md` ("Architecture at a glance") and `Model/README.md`. These have already drifted from the implementation.

### Proposed solution

Use [Mermaid](https://mermaid.js.org/) diagrams defined inline in Markdown (` ```mermaid ` fenced blocks), which GitHub renders natively. This makes architecture diagrams:
- **Version-controlled and diff-able** — changes show up as text in the PR that causes them.
- **Editable in-repo** — no external design tool or binary re-export.
- **Reviewable** — a diagram update can be required as part of a model-changing PR.

## Scope

- [ ] Replace the `<img>` in `Model/README.md` with a Mermaid flowchart reflecting the current pipeline 
- [ ] Decide whether `README.md`'s "Architecture at a glance" keeps a lighter overview diagram or links to the detailed one in `Model/README.md`.
- [ ] Optionally retire `Media/auto_e2e_architecture.jpg` once no docs reference it.
- [ ] Add a short contributor note (in `CONTRIBUTING.md` or `Model/README.md`) that architecture changes should update the Mermaid source in the same PR.

## Acceptance criteria

- The diagram matches the current `AutoE2E.forward` data flow.

apgeorg:
@m-zain-khawaja what do you think? Maybe there are other tools we could use!?

riita10069:
<img width="4368" height="1548" alt="Image" src="https://github.com/user-attachments/assets/23af2e10-3c28-43f1-903d-a2345ecf070f" />

riita10069:
I generated it automatically, does it make sense?

m-zain-khawaja:
I think this looks pretty good - happy to go with this 👍

apgeorg:
> I generated it automatically, does it make sense?

Do you generate an actual file, or are you referring to a code-based diagram? Mermaid diagrams are defined as code and rendered automatically by GitHub

riita10069:
I generated it using json and svg, I think mermaid is too difficult to improve apperance.
so more flexible tools are more good for this usecase like three.js or something.
I will make PR about this tool at /tools/vizualizer and make it easy to use by Makefile

apgeorg:
[...]

PR #90, opened 2026-06-24
Feat/import visual ml model

riita10069:
#70 

transplant from my personal repo to /tools


riita10069:
The remaining `lint-and-test` failure (`model_components.auto_e2e does not have the attribute 'Backbone'`) is caused by `Model/tests/conftest.py:94`, which patches `model_components.auto_e2e.Backbone` — an attribute that no longer exists in `auto_e2e.py`. This is a pre-existing bug on `main` itself (the latest CI runs on `main` are all failing as well), unrelated to this PR, which only adds files under `tools/`.

This PR changes nothing outside `tools/visual-ml-model/` and has no effect on the model code or its tests, so I am treating the test failure as out of scope here and will address it separately. DCO and lint (ruff) both pass. Merging.

Issue #86, opened 2026-06-23
Refactor: Simplifying the model codebase and removing modules


m-zain-khawaja:
# Description

As our research efforts have developed we have implemented many features and we need to prune and streamline the model.

## Key updates

- Remove Concat and Cross Attention Fusion and adopting BEV Fusion as our standard multi-view fusion method as it aligns best with other SOTA methods, and seamlessly integrates with encoded map features and also handles multiple camera mounting positions, enabling our system to flexibly adapt to different vehicle types
- Remove GRU planner since it is quite memory consuming and slow, retaining Flow Matching Planner and Bezier Planner
- Remove ego_hidden output from the planners, the World Action Model (slow reasoning) part of the architecture should be responsible for visual history creation and future state prediction
- Refactoring the main AutoE2E class to create the Reactive_E2E model and adding a placeholder for the World_Action_E2E model
- Updating the run_forward_pass script in the inference folder to account for the new model updates
- Removing loss calculation from the planner module since we already have a trajectroy planner loss and all planners output the trajectory in an identical format 

@riita10069 @gcordova10

m-zain-khawaja:
Changes have been pushed to main

PR #94, opened 2026-06-25
fix(tests): repair the unit-test suite after the Reactive_E2E refactor (#88)

### LINKED ISSUE #88: [Bug]: CI is broken because tests are for old architecture and need to be updated
### What happened?

Unit tests are broken after the refactor of the AutoE2E class since everything has been moved inside the Reactive_E2E module

### Steps to reproduce

```shell
make test
```

### Traceback / logs

```shell

```

### Device

CPU

### Local test status

Tests are failing locally

### Environment

_No response_

### Model config (if relevant)

_No response_

### Confirmation

- [x] I searched existing issues and did not find a duplicate.

gcordova10:
### Summary

Closes #88. The #86 refactor (everything moved inside `Reactive_E2E`) changed the model's public contract, but the unit tests still asserted the **old** API, so `make test` was red. This PR brings the whole suite back in line with the refactored architecture and fixes two pre-existing source-level `mypy` errors that also kept CI red.

No production behaviour changes beyond two type-only annotations — this is a test-and-typing repair.

### Root cause (what the refactor changed)

After #86:

- `AutoE2E` is a thin wrapper around `self.Reactive_E2E = ReactiveE2E(...)`; the `Backbone`, `FeatureFusion`, `MapEncoder`/`MapBEVFusion`, `TemporalMemory` and `TrajectoryPlanner` now live **under `Reactive_E2E.*`** (so parameter names are prefixed `Reactive_E2E.`).
- `AutoE2E.forward(...)` returns **only `trajectory`** `[B, num_timesteps*num_signals]`, in every mode. The old 3-tuple `(loss/trajectory, ego_hidden, future_features)` is gone.
- The planner `forward()` is inference-only and returns **just the trajectory** (no `ego_hidden`); the per-planner `compute_planner_loss` was removed (the training objective lives in the training loop).
- View fusion: only **`bev`** is registered now (`FUSION_REGISTRY == {"bev"}`); `concat`/`cross_attn` are gone. The `fusion_mode` kwarg was dropped from `AutoE2E`/`ReactiveE2E` (it is hard-wired to `"bev"` internally); `FeatureFusion` itself still accepts a `fusion_mode` arg.
- The **GRU** planner was removed; `PLANNER_REGISTRY == {flow_matching, bezier}`. `AutoE2E`'s default `planner_mode` is **`bezier`**. (Note: `ReactiveE2E.__init__` still carries a stale `planner_mode="gru"` default that would raise if it were ever constructed directly with defaults — `AutoE2E` always passes `bezier`, so the live path is fine; flagging it as a latent source nit.)
- The `future_features` output was removed from the forward path. `FutureState` is **still constructed** inside `ReactiveE2E` but is **not called** by `forward`, so it produces no output and receives no gradient.

The tests were written against all of the above, so they failed en masse (collection-time and assertion-time).

### What changed, file by file

**Test suite (adapted to the new contract):**

- **`conftest.py`** — `_build_model_with_mock_backbone` patches `model_components.reactive_e2e.Backbone` (was `auto_e2e.Backbone`); always BEV at an 8×8 grid; `model`/`full_model` fixtures parametrised on `["bev"]` only.
- **`test_auto_e2e.py`** — `model(...)` now returns a single `trajectory`; removed the `ego_hidden`/`future_features` shape tests; train-mode calls now return the trajectory (no 3-tuple); gradient/optimizer tests backward through a trajectory MSE and target `Reactive_E2E.{Backbone,FeatureFusion,TrajectoryPlanner}`, excluding the unused `FutureState` and the alpha-gated `MapEncoder`; `num_views` matrix is BEV-only.
- **`test_view_fusion.py`** — read fused features via `model.Reactive_E2E.{Backbone,FeatureFusion}`; registry/BEV-specific tests unchanged.
- **`test_feature_fusion.py`** — `fusion_mode="concat"` → `"bev"` (+ `view_fusion_kwargs={"bev_h":8,"bev_w":8}`).
- **`test_map_encoder.py`** — single-trajectory return; paths via `Reactive_E2E.{MapEncoder,MapBEVFusion}`; gradient checks now drive a trajectory-MSE backward. **`cross_attn` map fusion stays covered** (it is still a valid `map_fusion_mode`).
- **`test_history_encoder.py`** — single-trajectory return; `Reactive_E2E.TemporalMemory`; `one_hz`/`no_memory` paths verified via trajectory backward.
- **`test_causal_reasoning.py`** — asserts the new single-trajectory contract; the (independent) `CausalReasoningModule` is exercised on a 256-d embedding since the planner's internal ego embedding is no longer surfaced.
- **`test_bezier_planner.py`** — planner `forward` returns just the trajectory; removed tests for the deleted `compute_planner_loss`/`trajectory_loss`; default-planner test now asserts **bezier** (GRU removed); AutoE2E tests patch `reactive_e2e.Backbone` and use `Reactive_E2E.TrajectoryPlanner`.
- **`test_trajectory_planning.py`** — Flow-Matching `forward` returns the trajectory; removed the `compute_planner_loss` end-to-end / wrong-shape tests and `test_train_mode_requires_target` (the forward no longer requires a target); gradient test backprops through the integrated trajectory; AutoE2E-with-FM tests use the single-trajectory contract; "interchangeable planners" now compares **bezier vs flow_matching**.
- **`test_integration.py`** — `TestRealArchitectureSmoke` (runs in CI, `is_pretrained=False`) and the `@integration` tier adapted to the single-trajectory contract and `Reactive_E2E.Backbone`.

**Source (type-only `mypy` fixes, no behaviour change):**

- **`losses/trajectory_loss.py`** — class-level annotations `loss_fn: nn.Module` / `temporal_weights: torch.Tensor` so `mypy` resolves them instead of `nn.Module.__getattr__ -> Tensor | Module` (which flagged `self.loss_fn(...)` as "Tensor not callable").
- **`map_encoder/raster_map_encoder.py`** — bind timm's `feature_info` through `Any` to iterate it without a `union-attr` error.

### Tests removed (and why)

Tests for features deleted by #86 were removed rather than left asserting dead behaviour: `concat`/`cross_attn` **view** fusion, the **GRU** planner, the planner `compute_planner_loss`/`trajectory_loss`, and the `ego_hidden`/`future_features` returns. Everything still present (BEV fusion, bezier + flow_matching planners, residual + cross_attn **map** fusion, temporal memory, the reasoning head) stays covered.

### Testing

Run exactly as CI does:

```
ruff check                  # All checks passed
cd Model && mypy .          # Success: no issues found in 74 source files
pytest Model/tests          # 181 passed, 6 deselected
```

The 6 deselected are the `@integration` tier (real pretrained backbones), which `Model/pytest.ini` excludes from the default/CI run via `-m "not integration and not e2e_data"`.

### Notes

- `FutureState` is still constructed inside `ReactiveE2E` but is **not used** by its `forward`, so it receives no gradient; the gradient-coverage tests exclude it intentionally (flagging in case it should be wired in or dropped).
- Net diff is **−172 lines** (12 files, +317/−489) — mostly deleting expectations for removed features.

### DIFF
--- a/Model/model_components/losses/trajectory_loss.py
+++ b/Model/model_components/losses/trajectory_loss.py
@@ -5,13 +5,18 @@ import torch.nn as nn
 class TrajectoryImitationLoss(nn.Module):
     """Primary task loss: imitation loss over predicted trajectory."""
 
+    # Class-level annotations so mypy resolves these to their real types
+    # instead of nn.Module's ``__getattr__ -> Tensor | Module`` fallback
+    # (otherwise ``self.loss_fn(...)`` is flagged "Tensor not callable").
+    loss_fn: nn.Module
+    temporal_weights: torch.Tensor
+
     def __init__(self, loss_type: str = "smooth_l1", temporal_decay: float = 0.95,
                  num_timesteps: int = 64, num_signals: int = 2):
         # temporal_decay defaults to 0.95 so near-future predictions are
         # weighted more heavily than far-future ones; near-future accuracy
         # is more safety-critical for planning.
         super().__init__()
-        self.loss_fn: nn.Module
         if loss_type == "smooth_l1":
             self.loss_fn = nn.SmoothL1Loss(reduction="none")
         elif loss_type == "mse":
--- a/Model/model_components/map_encoder/raster_map_encoder.py
+++ b/Model/model_components/map_encoder/raster_map_encoder.py
@@ -4,6 +4,8 @@ Takes a BEV-space RGB map image and produces a spatial feature map at the same r
 features, so the two can be fused directly.
 """
 
+from typing import Any
+
 import torch
 import torch.nn as nn
 import timm
@@ -40,8 +42,11 @@ class RasterizedMapEncoder(nn.Module):
             in_chans=in_channels,
         )
  
+        # timm's FeatureInfo is exposed via __getattr__ (typed Tensor | Module),
+        # so bind through Any to iterate it without a mypy union-attr error.
+        backbone: Any = self._backbone
         self._feature_channels = [
-            stage["num_chs"] for stage in self._backbone.feature_info
+            stage["num_chs"] for stage in backbone.feature_info
         ]
         backbone_channels = sum(self._feature_channels)
  
--- a/Model/tests/conftest.py
+++ b/Model/tests/conftest.py
@@ -76,30 +76,30 @@ class MockBackbone(nn.Module):
         return self.backbone(image)
 
 
-def _build_model_with_mock_backbone(num_views, fusion_mode, device,
+def _build_model_with_mock_backbone(num_views, fusion_mode="bev", device=None,
                                     num_timesteps=64, map_fusion_mode="residual",
-                                    planner_mode="gru", planner_kwargs=None):
+                                    planner_mode="bezier", planner_kwargs=None,
+                                    **model_kwargs):
     """Construct AutoE2E with the mock backbone injected.
 
-    Patches Backbone at the module level during construction to avoid
-    loading pretrained weights entirely. Forces BEV fusion to a small
-    8x8 grid so tests stay fast and memory-light; the production default
-    (450x300) is exercised by dedicated configuration tests.
+    Post-refactor (#86): the model is ``AutoE2E`` -> ``Reactive_E2E`` and the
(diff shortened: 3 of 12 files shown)


riita10069:
@gcordova10 I really appreciate your contributions!

PR #78, opened 2026-06-21
feat(platform): MLOps training platform on EKS — Flyte + MLflow training loop (#14)

riita10069:
## Summary

Implements the **training infrastructure** tracked in #14 (Training Loop) as a
production **MLOps platform on AWS EKS**, orchestrated with **Flyte** and tracked
with **MLflow** — taking AutoE2E from "forward pass on random tensors" to a real,
reproducible **data → train → evaluate** pipeline running on GPUs in the cloud.

Closes #14. Related: #16 (dataset/DataLoader), #13/#15 (losses), #30 (frame loading).

---

## What #14 asked for vs. what this delivers

| #14 checklist | Delivered |
|---------------|-----------|
| DataLoader for sequential frame data | `data_processing` pre-extracts aligned frames → WebDataset shards; `make_pre_extracted_loader` streams them (no video decode at train time, fixes #30) |
| Optimizer (AdamW) | ✅ AdamW in `train_il` |
| LR scheduler | ✅ cosine (recorded in config/metadata) |
| Basic loop: forward → loss → backward → step | ✅ real `AutoE2E` + `TrajectoryImitationLoss` |
| Gradient clipping | ✅ `grad_clip` arg |
| Logging (TB/WandB) | ✅ **MLflow** (params, metrics, artifacts, model registry) |
| Checkpointing | ✅ checkpoint + full metadata as Flyte/​S3 artifacts |
| Mixed precision (AMP) | ✅ bf16 autocast + GradScaler |

Plus the surrounding production concerns #14 implies: dataset ingestion, evaluation,
offline-RL refinement, experiment tracking, secrets, and full IaC.

---

## Architecture

```
data_ingest ─▶ data_processing ─▶ train_il ─▶ evaluate_il_policy   (MLflow: imitation-learning)
 (HF download)  (frame extract,      (AutoE2E IL)       │
                 WebDataset shards)                     ▼
                                          train_offline_rl ─▶ evaluate_rl_policy  (MLflow: offline-rl)
                                              (IQL refine)
```

- **Orchestration**: Flyte on EKS **Auto Mode** (pure — no managed node groups).
- **Each task has its own Docker image** (`data-prep`, `training`, `eval`, `offline-rl`),
  built via in-VPC CodeBuild and pushed to ECR.
- **Data passing**: FlyteDirectory/FlyteFile over S3.
- **GPU**: `g6e.4xlarge` (L40S) via a Karpenter NodePool pinned to an ODCR-backed AZ.
[...]

Issue #66, opened 2026-06-16
Proposal: Open-loop evaluation pipeline (validation split + trajectory metrics)


riita10069:
## Background

The training loop is now functional (#14) and the model learns on real data — trajectory imitation loss decreases reliably on both L2D and NVIDIA PhysicalAI datasets. The next step is measuring *what* the model has learned against established metrics so we can compare with published baselines and track progress.

AutoE2E outputs `(acceleration, curvature) × 64 timesteps` at 10 Hz (6.4s planning horizon). This proposal covers how to evaluate these predictions using the metrics standard in the E2E autonomous driving literature.

## 1. Trajectory Integration: (a, κ) → (x, y)

Since evaluation metrics operate in displacement space (meters), we need to integrate the model's action-space outputs into position trajectories. The standard approach is forward Euler integration with a kinematic bicycle model:

```python
x, y, theta, v = 0.0, 0.0, 0.0, v_initial  # ego-centric frame
positions = []
for t in range(T):
    x += v * cos(theta) * dt
    y += v * sin(theta) * dt
    theta += v * kappa[t] * dt  # yaw_rate = v * kappa
    v += a[t] * dt
    positions.append((x, y))
```

Where `v_initial` is the ego's speed at the current frame (available from `egomotion_history[-4]` for L2D, or the egomotion state for nvidia). `dt = 0.1s` (10 Hz).

This is the same integration used by [VADv2 (2024)](https://arxiv.org/abs/2402.13243) and other action-space planners. The gap between waypoint-based and action-space models in benchmarks is explicitly discussed in ["Addressing the Waypoint-Action Gap" (2025)](https://arxiv.org/abs/2602.06214).

## 2. Primary Metrics (Open-Loop)

Following the nuScenes planning benchmark established by [ST-P3 (ECCV 2022)](https://arxiv.org/abs/2207.07601) and adopted by [UniAD (CVPR 2023)](https://arxiv.org/abs/2212.10156), [VAD (ICLR 2024)](https://arxiv.org/abs/2303.12077), and [SparseDrive (2024)](https://arxiv.org/abs/2405.19620):

### L2 Displacement Error (ADE)

```
L2_t = || (x̂_t, ŷ_t) - (x_t, y_t) ||₂
ADE@Hs = (1/Hs) × Σ_{t=1}^{Hs} L2_t
```

Reported at horizons: **1s, 2s, 3s, 6.4s** (extending the standard 3s to our full 64-step horizon).

[...]

PR #97, opened 2026-06-29
feat(evaluation): comfort + off-road metrics, baselines, val splits (#66)

gcordova10:
## What this adds (complements the existing eval, doesn't change it)

`Model/evaluation/` already provides `compute_open_loop_metrics` (multi-horizon
ADE/FDE + signal MAE) and `gate_check` (thanks @riita10069, via #78). This PR
adds the **remaining pieces of #66** on top, without touching those functions:

- **`metrics.py`**
  - `compute_comfort_metrics(pred_accel, pred_curv, initial_speed)` — longitudinal
    jerk, lateral acceleration, lateral jerk and yaw rate derived from the `(a, κ)`
    outputs, compared against **nuPlan comfort thresholds**, plus a
    `comfort_violation_rate`. Needs **no ground truth**.
  - `offroad_rate(positions, drivable_mask, meters_per_pixel)` — an off-road proxy
    for collision rate using the BEV **drivable mask**, since L2D has no
    other-agent labels.
- **`baselines.py`** — training-free `constant_velocity_baseline` and
  `hold_last_action_baseline` in the `(accel, curv)` action space, so they're
  scored by the **existing** `compute_open_loop_metrics`. These check whether
  perception actually beats ego-status extrapolation (the nuScenes ego-status
  critique, Zhai et al. 2023).
- **`splits.py`** — `geographic_holdout_split` (reserve whole cities, leakage-safe;
  Lilja et al., CVPR 2024) and a simple `episode_range_split` fallback.

## Notes

- Pure-numpy, additive, opt-in: `compute_open_loop_metrics` / `gate_check` and
  their exports are unchanged; this only appends new functions and two small
  modules.
- The displacement metrics from my earlier draft were intentionally **dropped**
  as redundant — `compute_open_loop_metrics` already covers ADE/FDE.
- 10 tests (`test_evaluation_complementary.py`); ruff + mypy clean.

@riita10069 since this sits directly on top of your evaluation module, happy to
fold any of it into your structure / Flyte workflow however you prefer.

Addresses #66 (comfort, off-road, baselines, splits).

### DIFF
--- a/Model/evaluation/__init__.py
+++ b/Model/evaluation/__init__.py
@@ -1,3 +1,27 @@
-from .metrics import compute_open_loop_metrics, gate_check, integrate_trajectory
+from .metrics import (
+    COMFORT_THRESHOLDS,
+    compute_comfort_metrics,
+    compute_open_loop_metrics,
+    gate_check,
+    integrate_trajectory,
+    offroad_rate,
+)
+from .baselines import constant_velocity_baseline, hold_last_action_baseline
+from .splits import episode_range_split, geographic_holdout_split
 
-__all__ = ["compute_open_loop_metrics", "gate_check", "integrate_trajectory"]
+__all__ = [
+    # existing (open-loop displacement metrics + gate)
+    "compute_open_loop_metrics",
+    "gate_check",
+    "integrate_trajectory",
+    # complementary: comfort + off-road (#66 §2-3)
+    "compute_comfort_metrics",
+    "COMFORT_THRESHOLDS",
+    "offroad_rate",
+    # training-free baselines (#66 §5)
+    "constant_velocity_baseline",
+    "hold_last_action_baseline",
+    # validation splits (#66 §4)
+    "episode_range_split",
+    "geographic_holdout_split",
+]
--- a/Model/evaluation/baselines.py
+++ b/Model/evaluation/baselines.py
@@ -1,0 +1,45 @@
+"""Open-loop evaluation baselines (#66 §5).
+
+Simple, training-free baselines in the model's action space ``(accel, curv)``,
+so they can be scored with the same metrics as AutoE2E. Their purpose is to
+reveal whether the perception pipeline contributes **beyond ego-status
+extrapolation** — if the model barely beats constant-velocity, perception isn't
+helping yet (the ego-status critique of nuScenes planning, Zhai et al. 2023).
+
+The trained "ego-status MLP" baseline from the proposal needs training and is a
+follow-up; these are the zero-training references.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+
+
+def constant_velocity_baseline(batch_size: int,
+                               num_timesteps: int = 64) -> tuple[np.ndarray, np.ndarray]:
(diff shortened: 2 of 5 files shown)


riita10069:
Thanks for the PR / great contribution. 

 I would like to confirm a few points.

1. Comfort metrics: the current implementation is described as using “nuPlan comfort thresholds”, but it only checks longitudinal jerk, lateral acceleration, “lateral jerk”, and yaw rate. In nuPlan, comfort also includes longitudinal acceleration bounds, yaw acceleration, and jerk magnitude; also the 8.37 threshold corresponds to max jerk magnitude, not lateral jerk. Could you either rename this as a nuPlan-inspired proxy, or extend it to match the nuPlan comfort definition more closely?

2. Off-road metric: this currently checks only the trajectory center point against the drivable mask. That is useful as a lightweight proxy, but it is not equivalent to drivable-area compliance because the ego footprint/corners can leave the drivable area while the center remains inside. Could you make that add ego footprint / mask dilation support?

3. Flyte integration: this PR adds useful functions, but the current Flyte evaluation path still computes ADE/FDE manually and logs only `eval/ade`, `eval/fde`, and `eval/gate_pass`. I think it's ok to separated PR, but I want to support Flyte workflows and I will give you a permission to acess flyte UI. this is because flyte ui is so easy to train and compare.

4. Splits: `geographic_holdout_split` is a good primitive, but it is not connected to actual L2D metadata yet. Please clarify that this is a helper that requires episode-level city labels, or add the metadata extraction path. Also, `episode_range_split` should guard against `num_episodes <= 1` and avoid empty train splits.

Overall, I think this is a useful addition, but I would not call it the complete resolution of #66 until the comfort definition and workflow integration are clarified.

gcordova10:
Thanks for the detailed review, @riita10069 — all four points were fair. I've pushed the fixes:

**Comfort.** Aligned to the full nuPlan `ego_is_comfortable` set: I added the longitudinal-acceleration bounds (+2.40 / −4.05) and yaw acceleration (1.93) that were missing. You were right about 8.37 — that's the jerk *magnitude* bound (√(lon² + lat²)), not lateral jerk, so I renamed it and now compute it correctly. The function reports a peak and a violation rate per bound, plus an overall comfort violation rate.

**Off-road.** Added an ego-footprint option: pass `ego_size=(length, width)` and it checks the four oriented corners (heading from `headings`, or from the travel direction if none is given) instead of only the centre, with an optional `dilation_px` margin from the boundary. The centre-only check stays as the lightweight default.

**Flyte.** Happy to keep that as a separate PR. Once I have the UI access you mentioned, I'll wire `compute_comfort_metrics` / `offroad_rate` into the workflow and log them next to `eval/ade`. Thanks for offering it.

**Splits.** `episode_range_split` now requires at least 2 episodes and never returns an empty train set. `geographic_holdout_split` documents that it expects caller-provided city labels — extracting those from L2D metadata is a follow-up — and raises if the holdout leaves either split empty.

I've also reworded the PR so it reads as complementary to the existing metrics rather than as closing #66.

riita10069 (approved):

PR #95, opened 2026-06-29
feat(data_parsing): L2D 1 Hz sequential multi-view windows for the World Model (#16)

gcordova10:
## What this adds

`L2DDataset` currently loads only the *current* frame and leaves `visual_history`
as zeros, with no future frames — so the World Model / JEPA (#13, #85) has neither
history nor targets. This PR adds **sequential 1 Hz multi-view windows** (the
"Sequential frame access (needed for #13)" checklist item in #16).

## Design (testable without `lerobot`)

- **`data_parsing/l2d/world_model_windows.py`** (new, pure / dataset-agnostic):
  - `stride_for_hz(source_hz, wm_hz)` — L2D 10 Hz → 1 Hz = stride 10.
  - `window_offsets(N, stride)` — past `[-(N-1)s..0]` (oldest→newest, current last)
    and future `[s..N*s]`.
  - `required_margins(N, stride)` = `((N-1)*s, N*s)`.
  - `build_windows(load_frame, row, ep_start, ep_end, num_frames=4, stride=10)` →
    `(history_frames, future_frames)`, each `[N, V, 3, H, W]`; raises `IndexError`
    if a window would cross the episode boundary. Takes a `load_frame(row)->[V,3,H,W]`
    callable, so it has no dataset/`lerobot` dependency and is unit-tested with a
    synthetic loader.
- **`data_parsing/l2d/dataset.py`** (opt-in, default OFF → byte-identical to before):
  - new args `include_world_model_windows=False, wm_num_frames=4, wm_hz=1.0, source_hz=10.0`;
  - `_load_multiview_frame(row)` refactor, reused for the current frame and every
    window frame;
  - `_build_sample_index` margins take the **max** of the egomotion window (64/64)
    and the World Model windows, so a valid frame always has a complete window;
  - `__getitem__` emits `history_frames` / `future_frames` when enabled;
    `L2DSample` gains the two optional keys (`NotRequired`, version-guarded import
    for py3.10/3.12).

## How it fits

- **#13 `train_il`**: `compute_step_loss` consumes exactly `batch["history_frames"]`
  / `["future_frames"]` → `encode_history → predict_future → jepa_loss` (these were
  synthetic before).
- **#85 World Model**: `history_frames` feeds `WorldActionModel`; `future_frames`
  are the JEPA targets.
- Does not overlap #93 (different area — `data_parsing`).

## 1 Hz / 10 Hz

[...]

Issue #30, opened 2026-06-08
Discussion: Training-scale frame loading strategy for NVIDIA PhysicalAI dataset


riita10069:
## Context

In #19, @bharatwrrr added the NVIDIA PhysicalAI dataset parser. During review, we identified that the current `load_camera_frame()` implementation reads the full video file into memory on every `__getitem__` call (via `video_path.read_bytes()` + `SeekVideoReader`). This is workable for forward-pass smoke tests but would be prohibitively slow and memory-intensive for actual training loops.

@bharatwrrr addressed metadata caching (egomotion DataFrames, camera timestamps) but left the video read path as-is due to TorchCodec serialization issues with multi-worker DataLoaders. We agreed to open a follow-up issue to explore the right approach.

## Problem

For a typical 20s clip at 30 FPS with 7 cameras:
- ~600 frames per camera × 7 cameras = 4,200 frames per clip
- But only ~72 valid sample points per clip (10 Hz egomotion with 64-step history/future windows)
- Current design: reads entire video bytes, constructs a SeekVideoReader, decodes 1 frame, then discards everything — repeated per camera, per sample

With `DataLoader(num_workers=4)`, this means multiple full video files resident in memory simultaneously.

## Options discussed in #19

1. **On-the-fly TorchCodec decoding** — Replace `SeekVideoReader` with TorchCodec's seek-based decoder. Faster per-frame access without reading full bytes. However, TorchCodec objects may not serialize cleanly across DataLoader workers.

2. **TorchCodec with cached frame-index mapping** — Pre-compute which frame indices correspond to valid sample points. At training time, only decode those specific frames. Avoids keyframe penalty for distant seeks.

3. **Pre-extracted aligned-frame cache** — An offline preprocessing step that extracts only the frames aligned to valid `(clip_uuid, sample_idx, camera_name)` tuples, saving them as individual images (JPEG/PNG) or a memory-mapped tensor shard. A manifest maps `(clip_uuid, sample_idx, camera_name)` → file path.

## Questions to consider

- What's the right balance between preprocessing complexity and training-time I/O speed?
- Should we support both "raw video" and "pre-extracted" modes, so contributors can run smoke tests without preprocessing but training uses the fast path?
- Is there a preferred on-disk format for cached frames (individual JPEGs, tar archives, HDF5, memory-mapped tensors)?
- How should we handle the case where a contributor only downloads a subset of clips — should preprocessing be incremental?

## Related

- #19 (original implementation)
[...]

PR #100, opened 2026-06-30
feat(data_parsing): offline pre-extraction of the World Model 1 Hz windows (#16)

gcordova10:
## What this adds

Training the World Model on real data is bottlenecked by **video decode**: the online `L2DDataset(include_world_model_windows=True)` (#95) builds each 1 Hz history/future window by decoding MP4 frames on the fly (~tens of seconds per sample). This PR adds an **offline pre-extraction** of those windows so training reads them at disk bandwidth instead.

It is **pure windowing logic** (no `lerobot`, no I/O baked in — decode and storage are injected, exactly like `build_windows` in #95), so it is unit-tested in isolation and adds nothing to the default training/inference path.

## Design (`data_parsing/preextract_world_model.py`, new)

Content-addressed, to avoid a ~`2·N·V`× blow-up: the 1 Hz windows of consecutive samples overlap heavily, so instead of storing `2·N·V` frames per sample, each needed 1 Hz frame of an episode is stored **once** and each sample keeps only the **index lists** for its window.

- `plan_episode_windows(episode_len, num_frames=4, stride=10)` → the valid sample indices, each sample's `(history, future)` 1 Hz index lists, and the **set of unique frames to decode once**. Reuses `window_offsets` / `required_margins` from `world_model_windows.py` (#95), so the pre-extracted windows are **identical** to the online ones.
- `extract_episode(load_frame, episode_len, save_frame, …)` → decode each unique frame once → `save_frame` sink.
- `assemble_window(frame_store, history_idxs, future_idxs)` → rebuild `[N, V, 3, H, W]` history/future at train time from a frame store (a mapping or a lazy callable), without decoding video.

## How it fits

Same windows as #95, but precomputed once: `history_frames` feed the `WorldActionModel` (#85) and `future_frames` are the JEPA targets (#13). Motivated by the frame-loading-strategy discussion in #30.

## Scope / not in this PR (kept honest)

This is the first of the small PRs outlined in #16 — **only the windowing logic + its tests**. The next two steps from that plan are deliberately left out to keep this reviewable:
- wiring it into `pre_extracted.py` so the WebDataset loader emits `history_frames` / `future_frames`;
- adding it to the Flyte `data_processing` workflow that writes the shards (with @riita10069, who owns that pipeline).

## Tests

`pytest Model/tests/test_preextract_world_model.py` → **10 passed** (planning ranges, content-addressed dedup, episode boundaries, and **equivalence with the online `build_windows`**); full suite 221 passed; ruff + mypy clean. Off current `main` (`ab3651e`).

### DIFF
--- a/Model/data_parsing/preextract_world_model.py
+++ b/Model/data_parsing/preextract_world_model.py
@@ -1,0 +1,119 @@
+"""Offline pre-extraction of the World Model 1 Hz multi-view windows (#16 / #30).
+
+The online ``L2DDataset(include_world_model_windows=True)`` builds the
+history/future windows by decoding video per frame (~tens of seconds per
+sample) — fine for correctness, far too slow for training. This module
+pre-extracts those windows **once, offline**, so training reads them at disk
+bandwidth (see ``data_parsing/pre_extracted.py``).
+
+Design — *content addressed* (avoids the ~2*N*V x blow-up):
+  Windows overlap heavily under 10 Hz -> 1 Hz subsampling, so we do NOT store
+  ``2*N*V`` images per sample. Instead, for each episode we store every needed
+  1 Hz multi-view frame **once** (keyed by its in-episode frame index), and each
+  training sample keeps only the *list of frame indices* for its history/future
+  window.
+
+This file keeps the planning/assembly logic **pure and dataset-agnostic** (unit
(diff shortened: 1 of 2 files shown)


m-zain-khawaja (approved):
Approve - thanks @gcordova10

PR #101, opened 2026-06-30
fix(dependency-conflict): updated the versions of dependencies to match lerobot 0.5.1

Arseni10Lk:
# Dependency version fix 

* Bumped up torch to 2.7.1 to still have cu118, from the original requirements but to also satisfy the lerobot requirement.
* Downgraded numpy to 2.2.6 since lerobot requires a version older than 2.3.0

A current setup is fully compatible with `lerobot`

# L2D forward pass fix

With the dependency version issues out of the way, I fixed the forward pass on live L2D data by correcting the signature when calling `model()`

This addresses and resolves the issue #71 
@m-zain-khawaja


Arseni10Lk:
@m-zain-khawaja 

Resolved the branch conflict; ready to merge

PR #106, opened 2026-07-03
feat(world action model): view attention pool to encode multi camera views for wam

### LINKED ISSUE #93: Proposal: World Action Model (1 Hz camera multi-view encoder without BEV view  + encoded features temporal fusion)
## Summary
This proposes a concrete, lightweight design for the World Action Model module, following the diagram's flow and discussion last meeting (Jun. 24) for the **World Action Model (WAM)** band:
following is from the latest design sketch by @m-zain-khawaja.
```
multi-camera images (7) @ 1Hz --> Vision Backbone (reuse, FROZEN) --> Feature Embedding
   --> Encoded Visual History --> Feature Decoding --> Visual Future Feature Prediction
   --> Future Feature Reconstruction Loss (JEPA)
```

My understanding of the update is that now the WAM's **Encoded Visual History** vector replaces the old BEV-derived `visual_history` (which is currently a zeros placeholder) and feeds two consumers: the **Reactive planner** and (later) the **Reasoning model**. The reconstruction loss is in **multi-view feature space, NOT BEV**.

This is a design sketch for discussion. Corrections and advice are very welcome.

## Design principles
- **Cheap.** The WAM runs at 1 Hz and only adds *small* trainable heads on top of a frozen backbone. Combine/compress/decode all operate on tiny token sets.
- **Reuse.** Lean on existing repo pieces: `backbone.py`, the per-view pooling in `FeatureFusion`, will copy the cross-attention pattern from `view_fusion`/`bev_fusion`, the GRU in `temporal_memory/one_hz_encoder.py` (or proposed a temporal attention-pool), and the `feature_reconstruction_loss.py` comparator.
- **Decoupled from the reactive trunk** (see backbone freezing below).

---

## Component 1  Vision Backbone (reuse + **FREEZE**)

Reuse the existing backbone (same arch/pretrained weights):
```python
x = camera_tiles.reshape(B * V, C, H, W)
features = self.Backbone(x)          # reuse but with requires_grad=False
```
**Freeze it.** Rationale (three wins):
1. **No conflicting gradients.** The backbone is also trained by `Reactive_E2E` for BEV + trajectory. If the WAM's JEPA loss also updated the backbone, the two objectives might fight (BEV/trajectory-optimal features vs. future-prediction-optimal features). Freezing removes that conflict. But we can also use a dedicated backbone copy for WAM (frozen) and Reactive (trainable) if we want to keep the two objectives separate.
2. **Stationary JEPA target (anti-collapse).** A frozen backbone gives a **fixed** target feature space, so the predictor can't trivially collapse (the BYOL/JEPA requirement for a non-trainable target encoder). The **same frozen backbone** encodes both the past (context) and the future (targets), no separate frozen copy / EMA needed.
3. **Cheap.** No backbone backprop; only the small WAM heads train.

Decision to confirm: **separate frozen copy (pretrained init)** vs **shared-with-stop-gradient** on Reactive's live backbone. Frozen pretrained copy is cleanest (stable targets, matches @m-zain-khawaja's frozen-init decision in #13); shared-stop-gradient is lighter but gives non-stationary targets as Reactive trains the backbone.

## Component 2  Feature Embedding (cheap cross-attention, BEV-style)

Per-frame, combine the V camera feature maps into **one per-frame embedding vector** via a **single-query cross-attention**. Thinking to adopt the `bev_fusion` learnable-query idea collapsed to one non-geometric query (no projection, since we're not in BEV), with learned camera embeddings borrowed from `CrossAttentionViewFusion`:

```python
class ViewAttentionPool(nn.Module):          # cheaper option: 1 layer, few heads
    def __init__(self, embed_dim=256, num_views=7, num_heads=4):
        super().__init__()
        self.view_embed = nn.Parameter(torch.randn(1, num_views, embed_dim))
        self.query      = nn.Parameter(torch.randn(1, 1, embed_dim))
        self.attn       = nn.MultiheadAttention(embed_dim, num_heads, batch_first=True)
        self.norm       = nn.LayerNorm(embed_dim)

    def forward(self, per_view_feat, B, V):   # per_view_feat: [B*V, C, h, w]
        C   = per_view_feat.shape[1]
        tok = per_view_feat.mean(dim=(2, 3)).reshape(B, V, C)  # pool each view --> [B,V,C]  
        tok = tok + self.view_embed                            # camera identity or learned positional embedding for the cameras
        q   = self.query.expand(B, -1, -1)
        out, _ = self.attn(q, tok, tok)                        # attend over V = 7 keys (asuming 7 cam)
        return self.norm(out.squeeze(1))                       # [B, C] per-frame embedding
```
Cost is negligible (attention over 7 tokens). Reuse `FeatureFusion`'s per-view multi-scale pool + `channel_proj` (1440-->256) to produce `per_view_feat`.

## Component 3  Encoded Visual History (compress the N-frame stack, cheap)

Maintain a **rolling stack of the last N per-frame embeddings** `[B, N, D]` (at 1 Hz). Compress to a single **Encoded Visual History** vector. Two options:

- **A. Temporal attention-pool (preferred --> consistent with Component 2).** A single learned query + **temporal positional embeddings** cross-attends over the N frame tokens --> `[B, D]`. This is the *same attention-pool pattern* as the view combiner, just applied over time axis instead of views, so the WAM uses one consistent primitive (cross-attention pooling) for both the spatial (multi-view) and temporal (multi-frame) aggregation. Cheap: 1 layer, few heads, attention over only N≈6 tokens.
  ```python
  tok = frame_embeddings + self.time_pos        # [B, N, D]  + temporal positional embeds
  q   = self.history_query.expand(B, -1, -1)    # [B, 1, D]
  out, _ = self.history_attn(q, tok, tok)       # attend over N≈6 frame tokens
  encoded_visual_history = self.norm(out.squeeze(1))   # [B, D]
  ```
- **B. GRU summary (reuse alternative).** Run a GRU over the N frame embeddings; final hidden = encoded history — the summarization half of `HistoryEncoder` in `temporal_memory/one_hz_encoder.py`. Marginally cheaper, but introduces a second aggregation primitive; kept as a fallback.

**Efficiency note:** because the backbone is frozen and the per-frame embedding depends only on its own frame, at 1 Hz you encode **only the new frame** each step and **cache** the previous N−1 embeddings. So per-step cost is one frame through the frozen backbone + the cheap compressor.

Output: `encoded_visual_history` `[B, D]` --> feeds the Reactive planner (replaces the zeros `visual_history`).

## Component 4 Feature Decoding (JEPA predictor, cheap & parallel)

Predict the **next N future per-frame multi-view features** (symmetric: `T_future = T_past = N`) from the encoded history. Use a **parallel query decoder** (no autoregression):

```python
# N learned future-step queries (+ temporal positional embeddings) cross-attend to the
# encoded history (or to the N past frame embeddings), then project to the frozen-backbone
# feature space (per-view pooled).
future_queries = self.future_pos.expand(B, -1, -1)            # [B, N, D]
dec, _ = self.decoder_attn(future_queries, context, context)  # 1-layer cross-attn
future_pred = self.to_feat(dec)                               # [B, N, V*C]  (per-view pooled features)
```
Cheap: one cross-attention layer + a linear projection. alternative: a GRU rollout for N steps (also cheap), but the parallel query decoder avoids sequential latency.

## Component 5  JEPA Loss

- **Targets:** the **frozen backbone** applied to the actual **N future frames**, per-view pooled --> `[B, N, V*C]`, **detached** (no grad). Because the backbone is frozen, targets are fixed -> no collapse, no EMA needed.
- **Loss:** MSE between `future_pred` and targets, optional temporal weighting (near-future weighted higher). **Reuse** `feature_reconstruction_loss.py`  this comparator (weighted per-step MSE) was **implemented by @gcordova10 in #80**, realising @riita10069's JEPA proposal (#13). We adopt it directly here — in practice, construct it with `num_future_steps=N` (its default is 4) and pass N per-step tensors (adapting the per-frame multi-view format).

```python
L_jepa = feature_reconstruction_loss(future_pred, target_future_feats.detach())
# Total: L = L_traj + λ * L_jepa
```

---

## I/O contract

```python
class WorldActionModel(nn.Module):
    def forward(self, multi_view_stream):    # [B, N, V, 3, H, W] @ 1Hz (past)
        ...
        return encoded_visual_history        # [B, D]  --> planner

    def compute_jepa_loss(self, multi_view_stream, future_frames):
        # future_frames: [B, N, V, 3, H, W] (frozen-backbone targets)
        return L_jepa, encoded_visual_history
```
Integration in `auto_e2e.py`: build `self.World_Action_Model`; in `forward`, run it on the 1 Hz stream --> `encoded_visual_history` --> pass into `Reactive_E2E` (replacing the input `visual_history`); surface `L_jepa` in train mode.

## Reuse map

| WAM piece | Reuse from |
|---|---|
| Vision backbone (frozen) | `backbone.py` |
| Per-view pool + channel_proj | `FeatureFusion` (pre-BEV path) |
| View cross-attention + camera embeddings | `view_fusion/bev_fusion.py` (learned query), `CrossAttentionViewFusion` (view embeds) |
| JEPA loss comparator | `losses/feature_reconstruction_loss.py` (implemented by @gcordova10, #80) |

## Key design decisions (to settle in the issue)

| Decision | Options | Lean |
|---|---|---|
| Backbone | separate frozen copy vs shared+stop-grad | frozen copy (stable targets) |
| Future-target granularity | pooled per-view vs full spatial maps | **pooled** (v1, cheap) |
| Temporal compressor | temporal attention-pool vs GRU (reuse) | **temporal attention-pool** (consistent with the view embedding) |
| Decoder | parallel query decoder vs GRU rollout | parallel (no latency) |
| `encoded_visual_history` dim | keep 896 (drop-in to planner) vs 256 (cleaner, needs planner `visual_history_dim` change) | TBD |
| N (= past = future frames) | e.g. 6 (6.4 s @ 1 Hz) | TBD |
| Action-conditioning | unconditioned vs condition future on planned trajectory (GAIA-style) | for now unconditioned for simplicity |

## Implementation plan (phased, additive)

1. **Module skeleton + synthetic test.** `WorldActionModel` with the I/O contract above, tested on random tensors (repo pattern), lands before data. Default off / zeros fallback so `Reactive_E2E` keeps working.
2. **Components 2-3** (embedding + compressor) --> produce a real `encoded_visual_history`.
3. **Components 4-5** (decoder + JEPA loss) with frozen-backbone targets.
4. **Wire into `auto_e2e.py`** behind a flag; surface `L_jepa` in train mode.
5. **Dataloader extension** (separate, ties to #16): emit the 1 Hz multi-view **past + future** stream `[B, N, V, 3, H, W]`.

## Open questions
- `encoded_visual_history` dim, keep 896 or move to 256 (planner change)?
- Pooled vs spatial future targets (cost vs richness)?
- Should the future prediction be **action-conditioned** (future views depend on where we drive)?
- N and the 1 Hz sampling, fixed or tunable per dataset?

@m-zain-khawaja I would be glad to contribute and start with component 2 and 3.

## References
- Architecture diagram (World Action Model @ 1 Hz band)
- #13 FutureState JEPA proposal (@riita10069); #80 feature-reconstruction loss + action conditioning (@gcordova10) — reused here; frozen-backbone target decision (@m-zain-khawaja, #13); #82 temporal memory / `visual_history` interface
- **Credits:** the JEPA reconstruction loss this proposal builds on was implemented by @gcordova10 (#80) from @riita10069's proposal (#13); the frozen-backbone target choice follows @m-zain-khawaja.
- JEPA (LeCun 2022); BYOL (Grill et al. 2020); MILE (Hu et al. 2022); GAIA-1 (Hu et al. 2023)

intisar1020:
### Motivation & Architectural Context

Currently multi-camera features were aggregated by taking a naive arithmetic mean across the view dimension (`.mean(dim=1)`) (related to #85). While computationally cheap, mean-pooling might  have the limitation of not being able to preserve the information of each view for spatial representations as the mean  will blends all camera views equally.

In this PR we introduce a flexible, parameterized **Cross-Attention Pooling** mechanism. By anchoring camera views with explicit slot-based positional embeddings, the network preserves the unique spatial signature of each camera angle, enabling the World Action Model (WAM) to learn direction-aware representations during self-supervised JEPA training.

#### 1. View Attention Pooling & Parameter Propagation

`num_views` is now propagated dynamically from the high-level `AutoE2E` architecture down through `WorldActionModel`, `FrameEncoder`, and `ViewAttentionPool`. This allows the learnable positional embeddings to adapt automatically to different sensor layouts (e.g., switching between 6-camera and 7-camera vehicle configurations) without code changes.

**`auto_e2e.py`**
```python
self.World_Action_Model_E2E = WorldActionModel(
    backbone=self.Reactive_E2E.Backbone,
    frame_embed_dim=visual_history_dim // history_len,
    history_len=history_len,
    num_views=num_views,
    **wmk,
)
```

#### 2. Single-View Fallback Optimization

When only one camera view is configured (`num_views == 1`), running a single token through a cross-attention block adds unnecessary computation overhead. In this case, the initialization pipeline automatically falls back to a lightweight "mean" aggregator (effectively a squeeze operation) and logs a warning so the behavior stays visible.

**`world_action_model.py`**
```python
if num_views == 1 and view_aggregator == "attention":
    logger.warning(
        "num_views is 1; falling back view_aggregator from 'attention' to 'mean' "
        "to avoid unnecessary attention overhead."
    )
    view_aggregator = "mean"
```

#### 3. Unit Test Updates
Added new cases to cover dynamic positional-embedding sizing and verify the single-view warning and fallback logic.

**`test_world_action_model.py`**
```python
def test_world_action_model_view_aggregator(device, caplog):
    m_mean = _wam(device, view_aggregator="mean")
    assert m_mean.encoder.view_aggregator == "mean"

    m_attn = _wam(device, view_aggregator="attention")
    assert m_attn.encoder.view_aggregator == "attention"
    assert m_attn.encoder.view_pool.view_embed.shape[1] > 0

    # Custom num_views configuration
    m_attn_6 = _wam(device, view_aggregator="attention", num_views=6)
    assert m_attn_6.encoder.view_pool.view_embed.shape[1] == 6

    # Fallback warning when num_views=1 and view_aggregator="attention"
    import logging
    with caplog.at_level(logging.WARNING):
        m_attn_1 = _wam(device, view_aggregator="attention", num_views=1)
    assert any("num_views is 1" in record.message for record in caplog.records)
    assert m_attn_1.encoder.view_aggregator == "mean"
    assert not hasattr(m_attn_1.encoder, "view_pool")
```

---

### Verification 

- **Core functionality:** Confirmed multi-view arrays aggregate correctly with no feature-map shape corruption.
- **Fallback behavior:** Verified via log capture that `num_views=1` correctly triggers the performance bypass.
- **Regression tests:** Confirmed downstream reactive-planner expectations for the `[B, 896]` history output dimension remain unaffected.

Closes #93

### DIFF
--- a/Model/model_components/auto_e2e.py
+++ b/Model/model_components/auto_e2e.py
@@ -39,10 +39,11 @@ class AutoE2E(nn.Module):
         if enable_world_model:
             wmk = dict(world_model_kwargs or {})
             history_len = wmk.pop("history_len", 4)
+            wmk.setdefault("view_aggregator", "attention")
             self.World_Action_Model_E2E = WorldActionModel(
                 backbone=self.Reactive_E2E.Backbone,
                 frame_embed_dim=visual_history_dim // history_len,
-                history_len=history_len, **wmk,
+                history_len=history_len, num_views=num_views, **wmk,
             )
             self.visual_history_buffer = RollingHistoryBuffer(history_len=history_len)
 
--- a/Model/model_components/world_action_model.py
+++ b/Model/model_components/world_action_model.py
@@ -24,11 +24,45 @@ loop. Reuses the merged building blocks (JepaTargetEncoder,
 FeatureReconstructionLoss).
 """
 
+import logging
 import torch
 import torch.nn as nn
 
 from .jepa_target_encoder import JepaTargetEncoder, compute_jepa_loss
 from .losses.feature_reconstruction_loss import FeatureReconstructionLoss
+
+logger = logging.getLogger(__name__)
+
+
+class ViewAttentionPool(nn.Module):
+    """Learnable multi-camera combiner: [B, V, C] -> [B, C].
+
+    Opt-in replacement for the equal-weight mean over views: a single learned
+    query cross-attends over the V per-view tokens (+ camera-id embeddings),
+    so cameras are weighted by relevance instead of averaged.
+    """
+
+    def __init__(self, embed_dim: int = 768, num_views: int = 7, num_heads: int = 4):
+        super().__init__()
+        self.view_embed = nn.Parameter(torch.randn(1, num_views, embed_dim) * 0.02)
+        self.query = nn.Parameter(torch.randn(1, 1, embed_dim) * 0.02)
+        self.attn = nn.MultiheadAttention(embed_dim, num_heads, batch_first=True)
+        self.norm = nn.LayerNorm(embed_dim)
+
+    def forward(self, view_tokens: torch.Tensor) -> torch.Tensor:
+        # view_tokens: [B, V, C]   (C = embed_dim = 768)
+        B, V, C = view_tokens.shape
+        # Handle dynamic view sizes by slicing or repeating positional embeddings
+        if V <= self.view_embed.shape[1]:
+            v_embed = self.view_embed[:, :V]
+        else:
+            extra = V - self.view_embed.shape[1]
+            v_embed = torch.cat([self.view_embed, self.view_embed[:, :extra]], dim=1)
+            
+        tok = view_tokens + v_embed
+        q = self.query.expand(B, -1, -1)
+        out, _ = self.attn(q, tok, tok)
+        return self.norm(out.squeeze(1))
 
 
 class FrameEncoder(nn.Module):
@@ -38,22 +72,43 @@ class FrameEncoder(nn.Module):
     """
 
     def __init__(self, backbone: nn.Module, feature_channels: int = 768,
-                 frame_embed_dim: int = 224):
+                 frame_embed_dim: int = 224, view_aggregator: str = "attention",
+                 num_views: int = 7):
         super().__init__()
         self.backbone = backbone
         self.proj = nn.Linear(feature_channels, frame_embed_dim)
+        if num_views == 1 and view_aggregator == "attention":
+            logger.warning(
+                "num_views is 1; falling back view_aggregator from 'attention' to 'mean' "
+                "to avoid unnecessary attention overhead."
+            )
+            view_aggregator = "mean"
+
+        self.view_aggregator = view_aggregator
+        
+        if view_aggregator not in ("mean", "attention"):
+            raise ValueError(f"Unknown view_aggregator {view_aggregator!r}. Available: 'mean', 'attention'.")
+            
+        if view_aggregator == "attention":
+            self.view_pool = ViewAttentionPool(embed_dim=feature_channels, num_views=num_views)
 
     def forward(self, frame: torch.Tensor) -> torch.Tensor:
         """``[B, 3, H, W]`` or multi-view ``[B, V, 3, H, W]`` -> ``[B, embed]``.
 
         For multi-view input each camera is encoded and the per-view embeddings
-        are mean-pooled into a single per-frame embedding.
+        are aggregated (via mean or attention) into a single per-frame embedding.
         """
         if frame.dim() == 5:                       # [B, V, 3, H, W]
             B, V = frame.shape[:2]
             feats = self.backbone(frame.reshape(B * V, *frame.shape[2:]))
             m = feats[-1] if isinstance(feats, (list, tuple)) else feats
-            m = m.mean(dim=(2, 3)).reshape(B, V, -1).mean(dim=1)  # GAP + pool views
+            m = m.mean(dim=(2, 3)).reshape(B, V, -1)             # [B, V, feature_channels]
+            
+            if self.view_aggregator == "attention":
+                m = self.view_pool(m)                            # [B, feature_channels]
+            else:
+                m = m.mean(dim=1)                                # [B, feature_channels]
+                
             return self.proj(m)                                  # [B, embed]
         feats = self.backbone(frame)               # [B, 3, H, W]
         m = feats[-1] if isinstance(feats, (list, tuple)) else feats
@@ -193,7 +248,8 @@ class WorldActionModel(nn.Module):
     def __init__(self, backbone: nn.Module, feature_channels: int = 768,
                  frame_embed_dim: int = 224, history_len: int = 4,
                  num_future_steps: int = 4, loss_type: str = "l1",
-                 history_aggregator: str = "concat", feature_hw: int = 8):
+                 history_aggregator: str = "concat", feature_hw: int = 8,
+                 view_aggregator: str = "attention", num_views: int = 7):
         super().__init__()
         if history_aggregator not in ("concat", "attention"):
             raise ValueError(
@@ -215,7 +271,10 @@ class WorldActionModel(nn.Module):
             if history_aggregator == "attention" else None)
 
         # Online per-frame encoder (shared backbone, trainable) -> 224 history embedding.
-        self.encoder = FrameEncoder(backbone, feature_channels, frame_embed_dim)
+        self.encoder = FrameEncoder(
+            backbone, feature_channels, frame_embed_dim,
+            view_aggregator=view_aggregator, num_views=num_views
+        )
         # JEPA target: a FROZEN, stop-gradient copy of the backbone FEATURE MAP
         # extractor (#1). Per Zain's #85 review (30/06) the JEPA reconstructs the
         # future backbone feature maps [B, C, hw, hw], not a pooled vector.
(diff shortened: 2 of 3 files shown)


m-zain-khawaja (approved):
Approved - thanks @intisar1020

PR #107, opened 2026-07-03
Make BEV fusion runtime-dynamic on num_views via a camera projection-operator ABI

### LINKED ISSUE #77: Proposal: Make BEV fusion runtime-dynamic on num_views to enable multi-dataset training of a single model
## Summary

To train a **single BEV model on multiple datasets with differing camera counts** (L2D = 7 views, NVIDIA PhysicalAI-AV = 8 views, Waymo = 5, nuScenes = 6, ...), `BEVViewFusion` must accept a **variable number of views (`V`) at runtime**, not a fixed `num_views` baked in at construction time.

The BEV design (#12, #35) is conceptually camera-count-agnostic: learnable BEV queries on a fixed `bev_h × bev_w` grid attend to whatever cameras are available, and the output is always `[B, embed_dim, bev_h, bev_w]` regardless of `V`. This is exactly what makes "produce a BEV feature from any dataset and train uniformly" possible.

However, the current implementation still couples several tensors to the constructor-time `self.num_views`, so a model built with `num_views=7` cannot consume an 8-view batch (and vice versa). Today this forces **one model per camera configuration**, blocking multi-dataset training.

## Where the coupling is (`Model/model_components/view_fusion/bev_fusion.py`)

The `forward()` already threads the runtime `V` correctly in most places (`reshape(B, V, ...)` at L213, `range(V)` loop at L238, `attn_weights.reshape(B, N, V, ...)` at L229). The problem is the remaining `self.num_views` / fixed-size parameters:

1. **L180–181** — `_project_to_2d` reshapes with `self.num_views`:
   ```python
   ref_2d      = ref_2d.reshape(B, self.num_views, N, num_z, 2)
   valid_depth = valid_depth.reshape(B, self.num_views, N, num_z)
   ```
   The upstream einsum (L162) is already `V`-dynamic via `proj`'s view dim, so this reshape should use the projection's view count, not `self.num_views`. When `V != self.num_views` this crashes.

2. **L83–84** — `attention_weights` is a fixed-width Linear:
   ```python
   self.attention_weights = nn.Linear(embed_dim, num_views * num_points_in_pillar)
   ```
   Its output is later reshaped with the runtime `V` (L229), so a model with `num_views=7` produces `7 * num_z` logits but L229 tries to reshape into `V=8` → shape mismatch. The attention weights must be parameterized over `num_points_in_pillar` only (per-camera, applied in the `range(V)` loop), independent of view count.

3. **L72–73** — `pseudo_projection` is a fixed `[num_views, 3, 4]` `nn.Parameter` fallback used when `camera_params is None`. For multi-dataset training, real `camera_params` are required anyway (each dataset has its own calibration), so the fallback's fixed view count is only a problem in the calibration-free path. We should either (a) require `camera_params` when `V != self.num_views`, or (b) make the fallback view-count-agnostic.

## Why this matters

- **Multi-dataset training** (#16, #62, #3): the repo already has parsers for L2D (7), NVIDIA (8), KIT Scenes, with Waymo (5) and nuScenes (6) on the roadmap. A unified BEV policy should learn from all of them, but differing `V` currently requires separate models.
- **Batching constraint stays**: a single mini-batch must still contain a uniform `V` (you cannot `torch.stack` a 7-view and an 8-view sample). Multi-dataset training therefore means **homogeneous-`V` batches alternating across datasets** — which a `V`-dynamic model handles naturally, but a `num_views`-fixed model cannot.

## Proposed change

Make `BEVViewFusion` fully runtime-`V` dynamic so one model instance can consume batches of any view count:

- L180–181: reshape using the projection's view dimension (derive `V` from `proj.shape[1]`) instead of `self.num_views`.
- L83–84: redefine `attention_weights` to output `num_points_in_pillar` per camera (computed inside the existing `range(V)` loop), removing the `num_views` factor. Camera-level weighting already comes from visibility-based averaging (`1/|V_hit|`), matching BEVFormer's SCA, so no fixed per-camera head is needed.
- `pseudo_projection`: require `camera_params` for the multi-dataset / `V != self.num_views` case; keep the fixed fallback only for the single-config calibration-free smoke-test path.

This is the natural completion of the camera-count-agnostic BEV design started in #2 (parametrize `num_views`) and #12 (BEV spatial cross-attention).

## Scope / non-goals

- **BEV mode only.** `concat` and `cross_attn` keep `num_views`-fixed Conv/Linear by design and remain single-dataset; multi-dataset training is a BEV-only capability.
- Mini-batches remain homogeneous in `V`; the DataLoader/sampler should group samples per dataset (each batch = one dataset's `V`). Cross-dataset mixing happens at the batch level, not within a batch.

## Acceptance criteria

- [ ] A single `AutoE2E(fusion_mode="bev")` instance runs a 7-view batch and an 8-view batch back-to-back without re-instantiation, producing `[B, embed_dim, bev_h, bev_w]` both times.
- [ ] Gradients flow to `bev_queries`, `value_proj`, `output_proj` for both view counts.
- [ ] Existing BEV tests (output shape, gradient flow, visibility masking) still pass.
- [ ] New test: alternating-`V` batches in one training loop step.

riita10069:
Closes #77.

## Summary

Makes `BEVViewFusion` fully runtime-`V`-dynamic so a single model instance can
consume batches of any camera count, and reworks the geometry interface from a
fixed `[B, V, 3, 4]` matrix into a differentiable camera *projection operator*.
Also separates the nav-map view out of the camera tiles and plumbs real camera
calibration end-to-end for datasets that publish it.

Per the #77 discussion, the geometry contract of a BEV encoder is not
fundamentally a pinhole `K @ T` matrix. BEVFormer's spatial cross-attention (Li
et al., ECCV 2022) only needs each BEV query mapped to sampling locations on
each camera image; whether that map is a linear pinhole matrix is an
implementation detail. Baking `[B, V, 3, 4]` in also cannot represent a fisheye
(f-theta) camera, which is the native model of the NVIDIA PhysicalAI-AV dataset.

## Projection-operator ABI (`view_fusion/projection.py`, new)

- `CameraProjectionModel` protocol: `project_ego_to_image(points_ego,
  image_transform) -> ProjectionResult(uv_norm, valid_mask, depth, metadata)`.
- `PinholeProjection` (the linear `K @ T` fast path; `from_KT(K, T)` keeps
  intrinsics/extrinsics separable so a rectified fisheye is expressed as a
  `rectified_pinhole`, not a faked pinhole), `FThetaProjection` (native fisheye,
  no rectification / no FOV loss), `PseudoProjection` (learnable calibration-free
  prior). `BEVViewFusion` never branches on camera model.
- `ImageTransform` carries the model-input frame so projection targets the tensor
  that enters the backbone (resize/crop) rather than the raw image.
- There is no `camera_params` matrix argument on any forward; a pinhole matrix is
  wrapped as `PinholeProjection(matrix)` by the caller.

## Runtime-`V`-dynamic fusion

- Removes the three constructor-time `num_views` couplings (`pseudo_projection`,
  `attention_weights`, the `_project_to_2d` reshape); `V` is derived from the
  projection at runtime.
- `attention_weights` drops the per-camera axis (per-pillar, shared across
  cameras); camera weighting stays visibility-based (`1/|V_hit|`), matching
  BEVFormer's SCA.

## Explicit, honest geometry

`forward` takes a `geometry_type` label consistent with the supplied operator.
The calibration-free pseudo path is entered only when explicitly requested,
never silently on behalf of a caller that meant to pass real calibration; a
dataset without calibration logs a warning rather than silently degrading.

## Map separation + real calibration

- The nav-map is not a camera (it has no projection), so it is routed to the
  existing map branch. L2D becomes 6 real cameras (was 7), NVIDIA 7 (was 8);
  shards store it under a distinct `map.jpg` key.
- NVIDIA calibration is ingested from the `physical_ai_av` SDK and built into a
  native `FThetaProjection`; it is a per-dataset rig constant stored once in the
  shard manifest and reconstructed by the loader.
- L2D has no published intrinsics, so it stays on the explicit pseudo path
  (follow-up); it is never labelled as real geometry.

## Acceptance criteria (#77)

- A single `bev` instance runs a 7-view and an 8-view batch back-to-back without
  re-instantiation, producing `[B, embed_dim, bev_h, bev_w]` both times: done.
- Gradients flow to `bev_queries`, `value_proj`, `output_proj` for both view
  counts: done.
- Existing BEV tests (output shape, gradient flow, visibility masking) pass: done.
- New alternating-`V` test: done.

## Testing

223 passed on GPU (g6e / L40S). New unit tests cover the projection operators in
isolation, the V-dynamic and honest-geometry contracts, native f-theta through
the fusion module, per-axis image normalization, device coercion, and the
shard map/camera split plus manifest projection round-trip. A full real-data
E2E smoke (NVIDIA clip -> shards -> loader -> forward -> loss -> backward) runs
green on both the pseudo and native-f-theta geometry paths. The projection code
was additionally reviewed by three independent tools; their findings (serialize
round-trip asymmetries, device-mismatch, per-view polynomial handling) were
fixed and covered by regression tests.

## Scope / follow-ups

- Batches remain homogeneous in `V`; cross-dataset mixing happens at the batch
  level, as noted in #77.
- L2D real calibration (unpublished intrinsics + reference-camera-to-ego
  transform) is a follow-up; L2D runs on the pseudo path until then.
- The f-theta projection is wired and unit-tested; its quantitative accuracy on
  real NVIDIA data (anisotropic-resize handling, extrinsic frame convention)
  still needs validation on downloaded clips and is flagged in-code.


m-zain-khawaja (approved):
Approved - thanks @riita10069

Issue #103, opened 2026-07-01
[Feature]: Reasoning band: encode horizon/region-aware confidence in the scenario vector


ZaynabEM:
### Problem / motivation

The Reasoning band (#98) currently plans to emit a scenario classification vector to the planner, but we haven't defined what that vector should actually encode beyond a scenario label. In parallel, during a review of a separate [VisionPilot 1.0](https://github.com/autowarefoundation/autoware_vision_pilot) HARA WP review, Michael Woon (Retrospect) shared a perception-safety framing that I think sharpens this design choice, and it lines up directly with the edge-case objective of #98 discussed today with @m-zain-khawaja.

Two ideas from it are relevant here:

1. Perception is not one monolithic function : it decomposes into Initiation → Detection → Termination, and the acceptable error profile depends on where in the horizon an object is (e.g. bias toward more Initiations far away, fewer Terminations near the ego path). In other words, error and confidence should be stratified by region/horizon and by sub-function, not collapsed into a single score.
2. Perception shouldn't have the final word : in a Perception → Prediction → Planning stack, the downstream planner should accept perception's known limitations and plan around them rather than trusting raw detections. This is exactly the intuition behind the zero-init planner gate that @gcordova10 proposed.

Today the reasoning vector risks being a flat scenario label with no notion of horizon or confidence, so the planner can't tell "confident, act normally" from "low visibility ahead, plan conservatively." That's the gap.
References: [Strategies for Safe Perception](https://www.retrospectav.com/blog/strategies-for-safe-perception) · [Solutions Before Standards](https://www.retrospectav.com/blog/solutions-before-standards) (Retrospect / Michael Woon).

### Proposed solution

Extend the C2 classification vector from #98 so it carries a horizon/region-aware confidence alongside the scenario class, and have the planner consume it as a limitation to plan around — not as a hard command.
Concretely, additive on top of what #98 already proposes (reusing the typed SceneContext from #81):

* **Stratified output**. Instead of one scenario label, the C2 vector encodes scenario class plus a coarse confidence banded by horizon region (e.g. near / mid / far, or along-path vs off-path), reflecting the Initiation–Detection–Termination split. Kept as a small typed vector (no free-form text into the planner).
[...]

PR #108, opened 2026-07-06
feat(reasoning): scenario-classification band with zero-init planner gate and per-horizon confidence (#98)

gcordova10:
Part of #98; the planner coupling and per-horizon confidence follow the convergence in #103.

## Why
This adds the trained **Reasoning band** from the 1 July design — the 1 Hz lane that classifies the driving scenario for the current + 4 future scenes and conditions the trajectory planner through a zero-init gate. A follow-up PR (#109) adds the offline teacher registry and the evaluation pipeline on top of this one.

## What this adds
- `reasoning/scenario_taxonomy.py` — the label space as an extensible multi-label registry: maneuver (7) / edge_case (6) / weather_env (8); new axes (e.g. KITScenes labels) append without breaking indices.
- `reasoning/reasoning_band.py` — `ReasoningBand`: trunk over the 896 Encoded Visual History → per-group sigmoid heads for current + 4 horizons, a **per-horizon confidence head** (#103, temporal-first) and a **`ZeroInitGate`** (FiLM-style, weights AND biases zero-initialised — the repo's alpha=0 pattern) that modulates the visual history fed to the planner. Strict no-op at init: enabling the band leaves the reactive baseline byte-identical until training moves the gate (tested).
- `reasoning/teachers/` — the swappable teacher interface (`VLMTeacher`) with a deterministic backend for CI; real VLM backends land in the stacked PR through the same registry.
- `training/losses/reasoning_loss.py` — the distillation loss lives **outside the model** (per the World-Model follow-up decision): BCE default plus an asymmetric-loss option ([2009.14119](https://arxiv.org/abs/2009.14119)) for the strong class imbalance.
- Wiring in `auto_e2e.py`: `enable_reasoning_band` (default **off**).

## Scope — NOT in this PR (next PR in the stack)
Offline teacher registry (Qwen2-VL / VideoLLaMA3 + cross-teacher agreement) · OpenAI-compatible teacher endpoint · long-tail split + intervention/faithfulness check · reasoning column in the speed benchmark — all in #109.

## Tests
`pytest Model/tests` → **309 passed** (+29 for this band). `ruff` + `mypy` clean.

### DIFF
--- a/Model/model_components/auto_e2e.py
+++ b/Model/model_components/auto_e2e.py
@@ -1,4 +1,7 @@
+from typing import Any, Dict, Optional
+
 import torch.nn as nn
+
 from .reactive_e2e import ReactiveE2E
 from .world_action_model import RollingHistoryBuffer, WorldActionModel
 
@@ -13,7 +16,9 @@ class AutoE2E(nn.Module):
                  map_fusion_mode="residual", map_fusion_kwargs=None,
                  temporal_memory_mode="no_memory", temporal_memory_kwargs=None,
                  planner_mode="bezier", planner_kwargs=None,
-                 enable_world_model=False, world_model_kwargs=None):
+                 enable_world_model=False, world_model_kwargs=None,
+                 enable_reasoning_band=False,
+                 reasoning_kwargs: Optional[Dict[str, Any]] = None):
         super(AutoE2E, self).__init__()
 
         # Reactive model which runs at 10Hz and processes multi-camera inputs
@@ -34,8 +39,8 @@ class AutoE2E(nn.Module):
         # future visual features (JEPA). Reuses the reactive backbone (one shared
         # backbone; the JEPA target is a frozen copy of it). Opt-in (default OFF)
         # so the reactive-only default is byte-identical.
-        self.World_Action_Model_E2E = None
-        self.visual_history_buffer = None
+        self.World_Action_Model_E2E: Optional[WorldActionModel] = None
+        self.visual_history_buffer: Optional[RollingHistoryBuffer] = None
         if enable_world_model:
             wmk = dict(world_model_kwargs or {})
             history_len = wmk.pop("history_len", 4)
@@ -45,6 +50,20 @@ class AutoE2E(nn.Module):
                 history_len=history_len, **wmk,
             )
             self.visual_history_buffer = RollingHistoryBuffer(history_len=history_len)
+
+        # Reasoning Band (slow, ~1Hz): classifies the current (and in training,
+        # future) scenario across multi-label taxonomy groups, supervised by
+        # the student/teacher loss, and feeds the trajectory planner through a
+        # ZERO-INIT gate that modulates the visual history (#98/#103) — a
+        # strict no-op at initialisation, so with the gate untrained the
+        # reactive baseline is unchanged.  Opt-in (default OFF) so the
+        # reactive-only baseline is byte-for-byte unchanged when disabled.
+        self.Reasoning_Band: Optional[nn.Module] = None
+        if enable_reasoning_band:
+            from .reasoning.reasoning_band import ReasoningBand  # local import — lazy
+            rkw = dict(reasoning_kwargs or {})
+            rkw.setdefault("visual_history_dim", visual_history_dim)
+            self.Reasoning_Band = ReasoningBand(**rkw)
 
     def reset_visual_history(self):
         """Clear the World Model's rolling buffer (call between sequences)."""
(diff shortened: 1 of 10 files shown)


riita10069:
@ZaynabEM Could you review this PR?

ZaynabEM:
@riita10069 just saw your comment thank you.
Will do as soon as I get the time

ZaynabEM (approved):
It's a solid, well-scoped first step toward the edge-case / safety objective in #103 . Flagging the follow-on work so we're aligned on what still discharges the full idea, but none of it blocks this merge.

### What's good and ready

- **Zero-init gate is exactly right.** `output = visual_history * (1 + gamma) + beta`, gamma/beta zero-init on weights and biases with strict no-op at init. This is the "perception shouldn't have the final word" property from #103 as a structural guarantee: we can add the whole edge-case lane and it *cannot* make the reactive baseline drive worse. That's the safety-by-design win (that I am trying to have), and it's the hardest part to get right.
- **Wiring is disciplined** : band sits after `aggregate_history`, before the planner; the only path to the planner is the gated tensor; default-off, direct import, byte-identical baseline preserved.
- **Temporal-first classification heads** are the right first move for edge-case *detection*, per what we agreed; stand up the scenario lane on the temporal axis first, layer region on after. Clean implementation (per-group, per-horizon heads; train gives all horizons, infer gives current).
- **Supervision stays train-only / student–teacher** (#103  point 4) is matched.

### Upcoming developments to fully discharge #103 

Flagging these as tracked next steps, not change-requests on this PR. The through-line: #103 's objective is that the planner *consumes* region-aware confidence "as a limitation to plan around", so the detection lane here becomes an actual safety function once the confidence signal is made load-bearing.

1. **Consume confidence in the planner coupling** (`reasoning_band.forward` / `ZeroInitGate`). Right now the gate reads the class probabilities (`current_probs`); `confidence` is returned but nothing reads it. Next step is to feed confidence into the gate so low confidence leads to more conservative modulation. This is the piece that turns "detects an edge case" into "drives more cautiously when it detects one."
2. **Supervise confidence** (`reasoning_loss.py`). It's currently an output with no target. Natural fit with #103 's plan: the teacher also labels a coarse per-region confidence (no new human labels). Cross-teacher agreement from the MultiTeacher direction (#109 ) is a good uncertainty label here. Worth pairing with a calibration metric (ECE) so the confidence is trustworthy before the planner leans on it; (1) and (2) should land together.
3. **Add the region axis** (near/mid/far, along-path/off-path): Temporal-first is the stepping stone; region is where most of the edge-case *safety* value lives (targeted caution where the hazard is, like the occluded pedestrian, vs. generic far-future caution). The hard part is the coupling: a single pooled visual-history vector can't easily express "be cautious only in that region," so this likely wants the gate to act on a spatially-structured BEV representation.
4. **Wire the "can I see?" channel** (#103  point 3): derive confidence partly from low-light / occlusion / clutter, or reuse the world model's reconstruction error as a proxy. Grounds confidence in a physical cause of uncertainty and connects to the self-belief / integrity-monitoring direction.

### Small thing to confirm

The band overwrites `visual_history` *after* `predict_future`, so the JEPA target sees the unmodulated history and the planner sees the modulated one. Probably intended (JEPA supervises the world model, not the reasoning-conditioned signal). Just confirm and maybe drop a one-line comment.

---

Approving as a clean incremental step. The containment guarantee is real and done, the detection lane is the right foundation. I'll open a tracked follow-up for items 1–2 (make confidence supervised + consumed, so the step that makes this actually improve safety) and a separate one for item 3 (region axis). 

Nice work @gcordova10 .

gcordova10:
Thanks @ZaynabEM, and thanks for the approval — really thoughtful review.

Confirming your observation: yes, it's intentional. The band overwrites `visual_history` **after** `predict_future`, so the World-Model / JEPA target is supervised on the *unmodulated* history (its own future-prediction task), while only the planner sees the reasoning-modulated history. That keeps the world-model self-supervision independent of the reasoning-conditioned signal — otherwise the JEPA target would be chasing a moving, reasoning-shaped input. I'll add a one-line comment at that point to make the ordering explicit.

And your follow-on items 1-4 are exactly right — that's the through-line from #103 to an actual safety function. I've already responded on the two issues you opened: on **#110** (supervise + consume confidence) the natural label-free target is this PR-stack's cross-teacher **agreement fraction** from `MultiTeacher` (#109), and consuming it as a monotone scaling of the gate residual preserves the zero-init no-op; on **#111** (region axis) the coupling likely wants to act on a BEV representation, reusing the World Model's BEV rather than a parallel path. Happy to take the #109-side hooks (expose the agreement signal + wire ECE into the eval module).

gcordova10:
Closing this one. The reasoning band landed in `main` through #98 — `HorizonReasoningHead`, the
compositional taxonomy, and the zero-init `ReasoningCoupling` — so there is nothing left here to
merge against.

The teacher-side half of the same arc went in separately as #109, so that part is in.

@ZaynabEM, thank you for the review back in July. It was a careful one and I appreciated the time
you put into it.

PR #109, opened 2026-07-07
feat(reasoning): VLM teacher registry + OpenAI-compatible endpoint + long-tail/faithfulness evaluation (#98)

gcordova10:
Part of #98; builds on the trained Reasoning band in #108 (this PR is stacked on that branch).

## Why
#108 adds the trained Reasoning band. This PR adds the **offline supervision and evaluation layer** around it: the swappable teacher registry that generates the band's pseudo-labels before training, a model-agnostic teacher endpoint so any OpenAI-compatible server (including @riita10069's Cosmos3-Nano vLLM PoC) can drive labelling, and the evaluation the thread asked for (long-tail stratification + an intervention/faithfulness check). Nothing here runs in the vehicle — every teacher is train-only.

## What this adds
- **`reasoning/teachers/`** — a swappable, offline teacher registry behind one `label()` interface:
  - `Qwen2VLTeacher` — working prompt builder + robust JSON parser over a closed label schema.
  - `VideoLlama3Teacher` — a video-native second opinion (the generation call is wired together with the GPU labelling pipeline; until then it raises a clear message instead of a silently wrong integration).
  - `MultiTeacher` — fuses teachers by per-label **agreement**, which doubles as the confidence signal, rather than trusting a single VLM's token probability.
  - **`OpenAIEndpointTeacher`** (new) — a **model-agnostic backend** that speaks the OpenAI chat-completions API and depends only on `base_url`, `model`, `prompt_version`, and the request/response schema. The network boundary is an injectable transport, so CI runs it with a stub (no GPU, no network).
- **`evaluation/`** — `long_tail_split` (stratify evaluation samples by rare classes) and `reasoning_intervention_delta` (a faithfulness check: does zeroing the band's planner coupling actually move the trajectory — 0.0 at initialisation by construction, so it also guards that enabling the band leaves the reactive baseline unchanged before training).
- **`speed_benchmark`** — a reasoning column measuring the band's added 1 Hz cost.

## Design notes (happy to adjust)
- The endpoint currently labels over the band's **current 3-axis taxonomy**; migrating the label space to the compositional action-relevant ontology is a follow-up — it grows through the same registry.
- It labels each horizon from its own frame (future frames as privileged offline information), rather than returning all horizons in one response. Both are valid; I can switch to a single-call horizons schema if preferred.

## Scope — NOT in this PR
The horizon-query cross-attention decoder, the pooled `reasoning_latent`, and the compositional action-relevant ontology are a separate follow-up PR — they build on this teacher/evaluation layer.

## Tests
`pytest Model/tests` → 370 passed, 4 skipped. `ruff` + `mypy` clean.

### DIFF
--- a/Model/evaluation/__init__.py
+++ b/Model/evaluation/__init__.py
@@ -7,7 +7,8 @@ from .metrics import (
     offroad_rate,
 )
 from .baselines import constant_velocity_baseline, hold_last_action_baseline
-from .splits import episode_range_split, geographic_holdout_split
+from .splits import episode_range_split, geographic_holdout_split, long_tail_split
+from .faithfulness import reasoning_intervention_delta
 
 __all__ = [
     # existing (open-loop displacement metrics + gate)
@@ -24,4 +25,7 @@ __all__ = [
     # validation splits (#66 §4)
     "episode_range_split",
     "geographic_holdout_split",
+    # reasoning band evaluation (#98/#103)
+    "long_tail_split",
+    "reasoning_intervention_delta",
 ]
--- a/Model/evaluation/faithfulness.py
+++ b/Model/evaluation/faithfulness.py
@@ -15,7 +15,7 @@ not perturb the reactive baseline before training.
 
 from __future__ import annotations
 
-from typing import Optional
+from typing import Any, Optional
 
 import torch
 
@@ -26,7 +26,9 @@ def reasoning_intervention_delta(
     map_input: torch.Tensor,
     visual_history: torch.Tensor,
     egomotion_history: torch.Tensor,
-    camera_params: Optional[torch.Tensor] = None,
+    projection: Optional[Any] = None,
+    geometry_type: Optional[str] = None,
+    image_transform: Optional[Any] = None,
 ) -> dict[str, float]:
     """Measure how much the reasoning band's coupling moves the trajectory.
 
@@ -36,15 +38,20 @@ def reasoning_intervention_delta(
 
     Args:
         model: an ``AutoE2E`` instance with ``enable_reasoning_band=True``.
-        camera_tiles / map_input / visual_history / egomotion_history /
-            camera_params: one evaluation batch, as in ``AutoE2E.forward``.
+        camera_tiles / map_input / visual_history / egomotion_history: one
+            evaluation batch, as in ``AutoE2E.forward``.
+        projection / geometry_type / image_transform: the current geometry ABI
+            forwarded to ``AutoE2E.forward`` (replaces the old ``camera_params``
+            argument).
 
     Returns:
         dict with:
         * ``trajectory_l2``: mean L2 distance between the coupled and
           intervened trajectories (0.0 while the gate is untrained).
-        * ``history_shift``: mean L2 between the modulated and raw visual
-          history (how hard the gate is actually steering the planner input).
+        * ``history_shift``: mean L2 between the modulated and the **effective**
+          visual history the band actually receives inside the model (with the
+          World Model on, that is the WAM-aggregated history, not the raw
+          caller input) — how hard the gate is steering the planner input.
 
     Raises:
         ValueError: if the model has no reasoning band to intervene on.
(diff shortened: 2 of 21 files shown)


riita10069:
@gcordova10  great work! I will review it.

riita10069 (commented):
Thanks for the PR. I like the overall direction: keeping teacher models offline/train-only, hiding the actual backend behind an OpenAI-compatible endpoint, and adding long-tail + faithfulness evaluation are all aligned with the #98 design direction.

I found a few issues that I think should be fixed before merging.

1. **`speed_benchmark` currently passes an unsupported `ReasoningBand` argument**

`run_speed_benchmark(..., reasoning="head_on_896")` builds:

```python
reasoning_kwargs={"backbone": reasoning}
```

but `ReasoningBand.__init__` does not accept `backbone`; it only accepts `visual_history_dim`, `hidden_dim`, `num_future_horizons`, and `taxonomy`.

So the reasoning benchmark path should fail with an unexpected keyword argument. I would either remove this kwarg and keep `reasoning` only as a result label, or add an explicit `reasoning_variant` config that `ReasoningBand` actually supports.

2. **The confidence head is currently untrained**

`ReasoningBand` defines and returns `confidence`, but `ReasoningLoss` only consumes `student_logits` and `teacher_targets`. There is no loss term for `pred.confidence`.

If confidence is part of the public output, we should train it. For example:

```python
target_confidence = agreement_fraction or source_weighted_label_confidence
L_confidence = brier_loss(sigmoid(pred.confidence), target_confidence)
```

Otherwise the confidence output should be removed or marked as placeholder, because right now it looks meaningful but is not supervised.

3. **Future horizons are predicted, but not actually used by the planner coupling**

The current gate is conditioned only on `logits[g.name][0]`, i.e. the current-horizon probabilities. The +1s to +4s predictions contribute to the auxiliary loss, but they do not directly condition the planner.

That weakens the main reason for horizon-aware reasoning: telling the planner *when* a hazard becomes relevant.

I think this PR can keep the current-only gate as the simple baseline, but the design should at least expose an ablation path:

```text
A. no reasoning
B. current-only pooled gate
C. all-horizon pooled latent
D. horizon-token cross-attention into the planner
```

Even if C/D are follow-up PRs, the output/interface should not make them harder.

4. **The OpenAI-compatible teacher is currently image-only and per-horizon independent**

I agree with the OpenAI-compatible endpoint abstraction. However, the current implementation sends one `image_url` per request and labels each horizon frame independently.

For Cosmos / video-native backends, I think we should support a second request mode:

```text
mode = "clip_horizons"
input = current + future frames, ego state, route/map metadata
output = one JSON object containing all horizons
```

The current `per_frame` mode is useful as a simple backend, but the endpoint abstraction should not be locked to single-image labeling. Otherwise we lose temporal context and cannot pass route/map/ego metadata, which are important for action-relevant labels.

5. **Please avoid silently converting endpoint failures into all-zero labels by default**

`OpenAIEndpointTeacher` catches all exceptions and returns empty labels. That is convenient for large jobs, but risky: endpoint outage, schema mismatch, or auth failure can silently poison the dataset with all-zero labels.

I would prefer:

```python
strict=True  # default, raise on endpoint/parser failure
strict=False # optional, abstain and record failure provenance
```

At minimum, the returned artifact should carry an `abstained` / `teacher_error` flag so downstream training can filter these samples.

6. **`MultiTeacher` should validate exact taxonomy compatibility**

`MultiTeacher` checks only `group_names`. It should also verify that each group has the same label tuple in the same order.

Otherwise two teachers with the same group names but different label order can produce semantically invalid agreement targets.

7. **`faithfulness.py` appears to use the old geometry argument and may measure `history_shift` against the wrong visual history**

The utility takes `camera_params`, but `AutoE2E.forward` now uses `projection`, `geometry_type`, and `image_transform`.

Also, when the World Model is enabled, AutoE2E replaces the input `visual_history` with the aggregated WAM history before applying the reasoning band. But `faithfulness.py` computes:

```python
pred = band(visual_history, mode="infer")
```

on the raw caller-provided `visual_history`, not necessarily the effective visual history used inside the model. So `history_shift` may not reflect the actual planner input.

I would update the utility to use the current projection ABI and to capture/compare the effective visual history that is actually passed through the reasoning band.

gcordova10:
Thanks @riita10069. Pushed the fixes in `1cc1722`; point by point:

**1. `speed_benchmark` passed an unsupported `backbone` kwarg.** Correct — `ReasoningBand` takes no `backbone`, so that path would have raised. Removed the kwarg; the band is now enabled with its defaults and `reasoning` stays purely a result label/column. (`speed_benchmark.py`)

**2. Confidence head untrained.** Agreed it shouldn't look meaningful while unsupervised. Added `confidence_brier_loss(confidence_logits, target_confidence)` in `reasoning_loss.py` — the proper-scoring-rule term you suggested — and marked `ReasoningPrediction.confidence` as a *trainable placeholder* in its docstring. The natural label-free target is exactly this PR's cross-teacher **agreement fraction** (`MultiTeacher`), so the full supervise-**and**-consume loop fits cleanly in your #110; happy to take the #109 side (expose the agreement signal + ECE in the eval module).

**3. Horizons not consumed by the planner.** Kept the current-only pooled gate as the baseline (your option B) and documented the **ablation surface** on `ReasoningPrediction`: `logits` exposes every horizon (now … +4 s), so an all-horizon pooled latent (C) or horizon-token cross-attention (D) can be built on the same output contract — the interface doesn't block them. C/D land with the horizon-decoder follow-up.

**4. Endpoint was image-only / per-horizon independent.** Added a second mode. `OpenAIEndpointTeacher(mode="clip_horizons")` sends the whole clip (current + future frames) in one request, with optional ego/route/map text via `extra_context`, and expects one JSON with a `horizons` array — so video-native backends (Cosmos) keep temporal context. `mode="per_frame"` stays as the simple backend. (Structured per-sample ego/map tensors would need extending the shared `label()` signature — flagged as a follow-up.) (`openai_endpoint.py`)

**5. Silent all-zero on failure.** Flipped the default to `strict=True`: a transport outage, auth error, or empty/malformed response now **raises** instead of poisoning the dataset with zeros. `strict=False` opts into best-effort abstain. When the Parquet artifact layer lands, abstained samples should also carry a `teacher_error` provenance flag for downstream filtering — noted as the next step. (`openai_endpoint.py`)

**6. `MultiTeacher` only checked group names.** Now it also verifies every teacher shares the **same label tuple in the same order** for each group, so index `i` is guaranteed to mean the same class before agreement fusion. (`multi_teacher.py`)

**7. `faithfulness` used the old geometry arg and the wrong visual history.** Both fixed: it now uses the current `projection` / `geometry_type` / `image_transform` ABI, and `history_shift` is measured against the **effective** visual history the band actually receives — with the World Model on that's the WAM-aggregated history, captured via a forward hook on the band, not the raw caller input. (`faithfulness.py`)

Gates after the fixes: `pytest Model/tests` → 383 passed, 4 skipped; `ruff` + `mypy` clean. Thanks again — this made the PR meaningfully better.

Issue #113, opened 2026-07-08
Evaluation reads the training shards — no train/val/test split (data leak)


riita10069:
@m-zain-khawaja

## Summary

`_run_evaluation` (Platform/pipelines/workflows.py) computes ADE/FDE on the
**same shards used for training**, so the open-loop metrics reported to MLflow
are optimistic (train-set fit, not generalization). There is no train/val/test
split anywhere in the data or training path yet.

## Where

- `evaluate_il_policy` / `evaluate_rl_policy` call `_run_evaluation(checkpoint,
  shards, ...)`, which does `shard_dir = _select_shard_dir(shards, dataset)` —
  the exact directory `train_il` trained on — and iterates its full WebDataset
  loader to compute ADE/FDE.
- `data_processing` / `wf_create_dataset` emit a single shard set with no split
  dimension; `make_pre_extracted_loader(split=...)` takes a `split` argument
  that is currently unused (all `.tar` in the dir are loaded).

## Impact

- Reported `eval/ade`, `eval/fde` measure fit to the training set; the gate
  check (`ade < 2.0 and fde < 4.0`) and any cross-run comparison (backbone,
  reasoning on/off, world-model on/off) are confounded by memorization.

## Proposed fix (open to guidance)

- Add a split dimension: pack `split=train|val|test` in
  `data_processing`/`wf_create_dataset`. `evaluation/splits.py` already has
  `episode_range_split`, `geographic_holdout_split`, and `long_tail_split`.
- `_run_evaluation` reads `val`/`test`, never `train`; thread the split through
  `wf_train_il` so train and eval consume disjoint shards.

## Question for the WG

Is there a dataset (or a specific split protocol) you'd recommend for
**evaluation** here? Options I can see:
- a held-out split of L2D itself (episode- or geographic-holdout);
- KITScenes-LongTail expert traces for a long-tail eval set (mentioned in #98);
- an external open-loop benchmark (nuScenes/Waymo-derived) if we want comparability.
Guidance on which held-out set best reflects the ODD we care about would shape
the split implementation.

## Note

This does not block the current smoke/wiring runs — a leaked eval is fine while
verifying the loss actually backpropagates and ADE decreases with training. It
must be fixed before any reported metric is treated as a generalization result.

Issue #110, opened 2026-07-07
[Feature]: Make reasoning-band confidence supervised and consumed by the planner (completes #103 safety loop)


ZaynabEM:
### Problem / motivation

#108 landed the reasoning band as a clean incremental step: temporal-first classification heads, the zero-init 
`ZeroInitGate` (byte-identical baseline, strict no-op at init), and train-only student/teacher supervision. That gives
us the **containment** guarantee — the edge-case lane cannot make the reactive baseline drive worse.

What it does **not** yet give us is the **active safety function** that #103 exists for. The band produces a `confidence` output, but:

- it is **not consumed**: the planner coupling (`ZeroInitGate`) is fed the class probabilities (`current_probs`), never `confidence`; and
- it is **not supervised**: `confidence` is returned from `ReasoningPrediction`  but has no target in `reasoning_loss.py`.

So confidence is currently observability at best. #103  's objective is that the planner *consumes* confidence "as a limitation to plan around"  i.e. detecting an edge case should make the car drive **more cautiously**. 

This issue closes that loop.

### Proposed solution

Two changes that must land together (acting on an unsupervised confidence would be worse than ignoring it):

**1. Consume confidence in the planner coupling** — `reasoning_band.forward` /`ZeroInitGate`
- Feed `confidence` (or `1 − confidence`) into the gate alongside  `current_probs`, or scale the gate residual by a monotone function of
  confidence, so **low confidence becomes more conservative modulation** of the visual history.
- Preserve the zero-init property: at init the gate must remain a strict no-op regardless of the confidence value.

**2. Supervise confidence** — `reasoning_loss.py`
- Give `confidence` a target. Options, roughly by rigor:
  - (a) against **correctness of the class prediction** (confidence as predicted reliability — a calibration target);
  - (b) against **teacher agreement** — cross-teacher disagreement from the MultiTeacher direction (#109) is a natural, label-free uncertainty signal;
  - (c) teacher-labelled coarse confidence (no new human labels, per #103).
- Add a **calibration metric (ECE)** to eval so we can tell whether the confidence is trustworthy before the planner leans on it.

### Possible Steps

- [ ] `confidence` feeds the planner coupling; low confidence demonstrably  produces more conservative modulation
- [ ] Zero-init no-op property preserved (test)
- [ ] `confidence` has a supervised target in `reasoning_loss.py`
- [ ] ECE (or equivalent calibration metric) reported in eval
[...]

PR #119, opened 2026-07-11
feat(#98/#13): 3-branch model (reactive + reasoning + WM) — end-to-end training, held-out eval, honest baseline

riita10069:
## AutoE2E: 3-branch camera-only driving model (reactive + reasoning + world-model)

Wires the full 3-branch architecture end-to-end through the Flyte pipeline and
makes it train correctly: Reactive planner (BEV → Bezier trajectory), the 1 Hz
Horizon-Aware Reasoning band (#98, Cosmos3-Nano teacher-distilled), and the
World-Model JEPA branch (#13). Design docs:
`Design/horizon_reasoning_architecture.md`, `Design/multi_view_fusion_architecture.md`.

### Architecture / design choices

- **Strictly-additive auxiliary branches.** Reasoning couples into the planner via
  a zero-init scalar gate (`alpha=0`); the World-Model's `visual_history` enters
  through a zero-init projection. At init the full model is identical to the
  imitation baseline, so the branches can only *add* value as they learn
  (regression-tested, `test_branches_strictly_additive.py`).
- **World-Model = JEPA on a stop-gradient backbone.** JEPA reconstructs future
  backbone feature maps; the shared trajectory backbone is stop-gradient'd in the
  WM encoder so JEPA cannot reshape the planner representation. The planner reads a
  detached world representation (standard world-model-conditioned policy).
- **Teacher as an independent, S3-cached stage.** `generate_reasoning_labels` calls
  Cosmos3-Nano once per sample and caches per-sample labels in S3 (keyed by
  dataset/teacher/prompt_version); `data_processing` JOINs them into shards. Never
  re-billed for an unchanged prompt. Temporal front-camera labeling via lerobot
  delta_timestamps (bit-identical to the full WM-window decode, far cheaper).
- **Held-out train/val split.** A deterministic per-sample hash of `__key__`
  partitions samples into disjoint train/val; eval scores the held-out `val` split
  so ADE/FDE measure generalization, not training-set memorization.
- **Gradient accumulation** recovers a larger effective batch when WM windows force
  `batch_size=1` (unit-tested identical to a full-batch step).

### Bugs found and fixed

1. **Eval train/eval `visual_history` skew.** Training fed the planner a dense
   WM-derived `visual_history`, but `_run_evaluation` never passed `history_frames`,
   so a WM-trained model ran the rolling-buffer path at eval (~3/4 zeros, out of
   distribution). Fix: feed the WM windows at eval → re-evaluating the same 30ep
   checkpoint dropped ADE 2.409 → 2.020 m with no retraining.
2. Non-additive WM input → zero-init `visual_history_proj`.
3. Trajectory loss leaking into the WM → detach `visual_history` for the planner
   (after `predict_future`, so JEPA still trains the WM).
4. Curvature loss scale ~9× off (0.014 vs measured 0.12) → corrected.
5. JEPA reshaping the shared backbone → `FrameEncoder.detach_backbone`.
6. Adversarial-review fixes: fp32 normalize before fp16 in the BEV projection;
   always-emit masked `reasoning__*` keys so mixed batches collate; OOV multi-label
   backfill with the abstain label; skip all-zero NVIDIA map tiles; label-cache hit
   counted only after successful deserialize.
7. **CI green.** ruff (unused imports), mypy strict (worker-global `Any`
   annotations, explicit `Optional`, dict kwargs typing), and `pytest.importorskip`
   for NVIDIA-only deps (pandas / physical_ai_av).

### Results — matched shard (L2D, 10 episodes / 1037 samples), lr=5e-4, 60 epochs

| Eval | Config | ADE | FDE |
|---|---|---|---|
| in-sample    | imitation-only | 1.299 m | 3.627 m |
| in-sample    | full 3-branch  | 1.407 m | 4.072 m |
| held-out 20% | imitation-only | 1.764 m | 5.783 m |
| held-out 20% | full 3-branch  | 2.218 m | 6.308 m |

(Progression from the bug fixes, full 3-branch @30ep in-sample: 2.409 → 2.020 →
1.860 m.) All branches train and converge; a one-time grad-flow probe confirms each
branch receives gradient. Held-out numbers are worse than in-sample for both models
— as expected, since the split measures generalization rather than memorization.

### Honest conclusion: at 10 episodes the reasoning/WM branches do NOT beat imitation

On this 1037-sample shard the full 3-branch model is consistently *slightly worse*
than the reactive-only imitation baseline, both in-sample and held-out. This is not
a bug or a measurement artifact — it is the correct behavior at this data scale:

- After 60 epochs the trained `reasoning_coupling.alpha` is still ~0.004 (from its
  zero init). The optimizer learned that the reasoning latent does not reduce
  trajectory loss here — expected, because the latent is a teacher-supervised
  re-encoding of inputs the planner already sees (visual_history / ego), and on
  1037 near-duplicate frames it carries no extra trajectory-predictive signal.
- The WM branch *did* open its projection (absmean 0.021), but its temporal signal
  is largely redundant with the 64-step egomotion history the reactive baseline
  already consumes, and its non-stationary features generalize slightly worse on so
  little data.

The architecture is designed to degrade gracefully to exactly this: the branches
are strict no-ops until they learn to help, so enabling them never has to regress
the baseline. Whether the reasoning band improves *held-out* ADE is a hypothesis
about data scale + scenario diversity — testing it needs 100+ episodes. Note
`yaak-ai/L2D` has 100,000 episodes and these runs use 10.

### Next steps

Scale to 100k L2D episodes (Cosmos labeling for those episodes is the gating cost)
and re-run the held-out Full-vs-imitation comparison. The zero-init coupling makes
this safe to attempt without regressing the reactive baseline.


riita10069:
## Detailed experiment log & analysis (follow-up)

### Full matched-condition results (L2D, 10 episodes / 1037 samples, lr=5e-4, fixed eval)

| Eval | Config | Epochs | ADE | FDE | final train loss |
|---|---|---|---|---|---|
| in-sample    | imitation-only | 30 | 1.771 | 5.157 | ~0.08 |
| in-sample    | full 3-branch  | 30 | 1.860 | 5.045 | 0.083 |
| in-sample    | imitation-only | 60 | 1.299 | 3.627 | 0.075 |
| in-sample    | full 3-branch  | 60 | 1.407 | 4.072 | 0.079 |
| held-out 20% | imitation-only | 60 | 1.764 | 5.783 | 0.078 |
| held-out 20% | full 3-branch  | 60 | 2.218 | 6.308 | 0.075 |

Bug-fix progression (full 3-branch, 30ep, in-sample): **2.409 → 2.020** (eval-skew fix) **→ 1.860** (zero-init proj + detach).

### Interpretation

- **The bug fixes were real and large** (full 3-branch ADE 2.409 → 1.860 m at 30ep). All three branches train and converge; a one-time grad-flow probe confirms nonzero gradient reaches planner / world-model / reasoning.
- **At 10 episodes the full 3-branch model does not beat the reactive-only imitation baseline** — in-sample (1.407 vs 1.299) and held-out (2.218 vs 1.764). This is the correct behavior at this data scale, not a regression:
  - Trained `reasoning_coupling.alpha ≈ 0.004` after 60ep (from its zero init): the optimizer learned the reasoning latent adds no trajectory-predictive signal here. The latent is a teacher-supervised re-encoding of inputs the planner already consumes (visual_history / 64-step egomotion history), so on 1037 near-duplicate frames it is redundant.
  - The WM projection did open (absmean 0.021), but its temporal signal overlaps the egomotion history the reactive model already has, and its non-stationary features generalize slightly worse on so little data.
- **The zero-init coupling means enabling the branches never has to regress the baseline** — they stay strict no-ops until they demonstrably help. That is exactly what we observe.

### Why held-out matters (a bug this surfaced — #113)

The original eval scored the *same shards used for training* (the loader ignored `split`). That is in-sample and structurally favors the lower-capacity model. The held-out split added here (disjoint deterministic per-sample hash) is the honest test; both models score worse on held-out than in-sample, confirming the split measures generalization. This closes #113.

### Conclusion on the hypothesis "reasoning/WM improve ADE"

It can be neither validated nor refuted at 10 episodes — the reasoning branch is a learned no-op because the data lacks the scenario diversity for teacher-distilled reasoning to carry generalizing trajectory structure. **Testing it requires data scale (100k episodes).**

Reaching that scale is a substantial parallelization effort across the whole pipeline (ingest / labeling / packing / training / eval) — today every stage is a single Flyte pod with intra-pod process pools. That work is split into a dedicated issue with per-item file:line evidence: **#121**. The single prerequisite there is replacing the positional `sample_id` with a global episode-anchored id so map-task fan-out keeps the label↔pack JOIN correct and the reasoning-label cache reusable.

gcordova10:
@riita10069 — great work here, and a few things deserve calling out specifically. The eval `visual_history` skew is the catch of the month: re-scoring the same checkpoint from ADE 2.409 to 2.020 with no retraining would have silently invalidated every number we reported afterwards. The held-out split matters for the same reason. And reporting the negative result — that the 3-branch model does *not* beat the imitation baseline at 10 episodes — instead of quietly dropping it is the right call.

Your root-cause in #121 is the sharpest thing written on this branch: if the reasoning head's inputs are a strict subset of the planner's, the latent cannot carry trajectory information the planner lacks, so `alpha → 0` is the optimizer being correct rather than a zero-init trap. That reframes the problem from "needs more data" to "needs a different input".

One practical thing I want to put on the table before this merges, plainly.

**We now have two implementations of the same band.** #119 adds a reasoning head, a taxonomy, a zero-init planner coupling, an offline teacher pipeline and a faithfulness module. #108 (approved) and #109 (open, with your 7-point review applied) add those same five things under different names. That is not a complaint — my PRs sat open, you needed a working pipeline to get numbers, and you got them, which we did not have. It just means we should land one band rather than maintain two, and I'd rather aim my time at what is still missing than at something you already have working. Easier to settle that now than after the merge.

**My proposal: #119 lands as the base.** It has the end-to-end wiring, the Cosmos3-Nano teacher with the label cache and Parquet artifacts, the split, and the results. I'll close or rebase #108/#109 onto it — tell me which you prefer and I'll do the work.

And to be explicit about one thing I offered earlier and have since checked: **the intervention/faithfulness metric is already yours, and yours is more complete than mine.** `evaluation/faithfulness.py` in this PR has `reasoning_intervention_delta`, and also `horizon_intervention_delta` with targeted per-horizon interventions (`zero_1s`, `zero_2s`, `shuffle`), which mine does not have. Use yours; there is nothing for me to port there.

**Three things I'd like to bring on top of this branch**, as small separate PRs:

1. **Teacher-embedding production for the alignment loss.** `HorizonReasoningLoss` already implements the alignment term and wires it correctly, but it is inert today: `lambda_alignment` defaults to `0.0`, and nothing in `reasoning_label_generation/` produces `teacher_embedding_targets` — `targets.py`, `teacher_client.py`, `openai_compatible.py` and `parquet_writer.py` have no notion of one. So `student_reasoning_embedding` is currently an untrained projection, and the dense signal it should align to is sitting unused: the teacher's per-horizon `evidence` rationale. I've built the missing link — a small offline module that turns that rationale into the `[B, 5, D]` target your loss already expects. Backend-agnostic (any sentence encoder), offline only, with a deterministic dependency-free encoder so CI needs no model download, and a validity mask so horizons without a rationale — and abstained records, per your R9 masking rule — are dropped rather than aligned against a zero vector.

   Two caveats I hit while checking it against the pipeline, which I'd rather put on the record than discover in review:

   - **The rationale is generated but not persisted.** `parquet_writer.py` keeps the provenance and the structured labels, not the text — so by the time training reads an artifact the rationale is gone. The embedding therefore has to be computed *during* label generation, where the teacher response is still in hand, and stored as its own column. Cheap (one encoder pass on text we already paid the teacher for), but it is a pipeline change, not a drop-in.
   - **Your schema marks `evidence` as "audit/debug only, never a training signal", and I want to check that with you rather than route around it.** It sits in tension with §6 of your #98 proposal, where the teacher reasoning embedding "comes from precomputed teacher text or teacher explanation features". My reading is that the schema note means "don't parse the free text into labels" — which would be brittle, and is what the structured fields are for — rather than "never derive a dense embedding from it", since an embedding is not a label but the soft semantic target your alignment term was built to consume. But you wrote both, so: **is embedding the per-horizon `evidence` the intended source for §6, or did you have a different one in mind** (a Cosmos-Embed vector over the clip, the teacher's hidden states)? If it's a different source, the module I built is the wrong shape and I'd rather know now than after a PR.

   This also unblocks something else: @intisar1020 and I were about to build OOD/drift checks on the `student_reasoning_embedding` tap in #98, and against an unaligned head those checks would be comparing a random projection to teacher prototypes — a number that looks fine and means nothing.

2. **Cross-teacher agreement as a label-free confidence target** (`MultiTeacher`): teachers fused per label by agreement fraction. It gives a calibrated confidence with no new labels, and it is what @ZaynabEM's #110 needs as its supervision signal.

3. **The unpooled-BEV reasoning input from your #121 root-cause** — I'd like to take this one if you're happy for me to. `HorizonReasoningHead` has no BEV input today, and it is your own prescribed fix for `alpha ≈ 0`. I have a horizon-query cross-attention decoder built against your #98 §2-§4 spec whose context tokens are already projected per source (visual, ego, with slots for route/map), so adding the BEV as one more context source drops into `horizon_reasoning_head` without changing its interface. One design note: the BEV grid is 450×300, so attending it cell-per-token is not viable — I adaptively pool it to a coarse grid (16×16, an ablation knob) with a 2-D position embedding, which keeps *where* the hazard is while staying cheap at 1 Hz. And I'll respect both of your caveats: two independent switches, so "the head sees the BEV" and "the reasoning loss reshapes the shared trunk" can be attributed separately (default: see-but-don't-reshape), and validated on an episode-count sweep rather than assumed — this repo has already found twice that auxiliary gradient into the shared backbone hurts at low data.

Tell me which split is most useful to you and I'll start there. Thanks for pushing this to real numbers; that was the thing we were missing.

gcordova10:
Following up on the three things I said I'd bring — **they're no longer proposals, they're PRs, open
against this branch rather than `main`:**

| | | |
|---|---|---|
| **#129** | `fix(data_ingest)` — NVIDIA's calibration never arrives; the run silently falls back to `geometry_type='pseudo'` | 1 file, +26 |
| **#130** | The unpooled BEV as a spatial context source — your #121 fix | 4 files, +256 |
| **#131** | Produce the `teacher_embedding_targets` the alignment loss expects | 3 files, +375 |

CI green on all three. `MultiTeacher` I'm holding — @m-zain-khawaja asked for no new features before the
benchmark, and I think that's right.

## #129, because one line undersells it

`get_clip_feature` is called without `maybe_stream=True` (the camera reads seven lines above do use it).
It raises, the `except` on `:207` turns that into a `WARN`, and the run continues on
`geometry_type='pseudo'` — a learned prior — while everything downstream assumes real f-theta.

**So NVIDIA, the only dataset that ships real calibration, has never actually received it, and the
f-theta path from #77/#107 has never run on real data.** Reproduced on a real 10-clip ingest: every clip
logged the WARN and the calibration directory was empty.

## #130, and why now rather than after the merge

First, a correction to what I said before: I told you I had the decoder built against your #98 spec.
**That was my own head, not yours** — measuring it would have answered a question about *my* architecture,
not the one that merges. So I rebuilt it on `HorizonReasoningHead`. It goes in as a third context source
next to the `route`/`map` slots you already left open, and fits **without touching the decoder**.

**#126 freezes the architecture on 17 July.** The reasoning branch as it stands here is — by your own
analysis — **a branch the model provably ignores** (`alpha ≈ 0.004`). Freezing it in that state means the
August benchmark runs on a three-branch model where one branch contributes nothing **by construction**,
and the natural read will be *"the reasoning branch doesn't help"* — which the data-processing argument
already tells us in advance, and which says nothing about whether the idea works.

**#130 is the input you prescribed for exactly that.** Measured on the head in this PR (RTX 3090, fp32,
WM off):

| Arm | Params | Peak, `bs=1` |
|---|---|---|
| imitation only | 91.35 M | 9.97 GB |
| reasoning, current inputs | 93.91 M | 9.98 GB |
| **reasoning + unpooled BEV** | **94.04 M** | **9.98 GB** |

**+0.132 M params (+0.14 %), under 10 MB, off by default** — and fixed *before* any run: **if the BEV arm
beats the no-BEV arm, it cannot be capacity. There is none to attribute. It can only be the
information.**

It also fixes one thing regardless of what's decided about the BEV: `train_il` never passes
`reasoning_kwargs` (`workflows.py:793-798`), so the head's `route`/`map` slots **cannot be turned on by any
training run in the repo** today.

---

Happy to rebase or split any of them. If you'd rather land #119 first and take #130 right after, that
works too — **the only thing I'd ask is that the reasoning branch isn't *frozen* without it**, since by
your own analysis that's the difference between a branch that's used and one that isn't.

riita10069:
Refactoring (AutoE2E Class has Reactive & World) will be worked by otner PRs. I shoud merge it because it takes us 2 days to train full dataset of KITScenes.

PR #125, opened 2026-07-14
Fix/116 camera loader oom

### LINKED ISSUE #116: NVIDIA camera loader reads each full mp4 into memory per camera every __getitem__ (bottleneck/OOM)
## Summary
`load_camera_frame` in the NVIDIA PhysicalAI-AV parser reads the entire clip mp4 into RAM (`io.BytesIO(video_path.read_bytes())`) for each of the ~7 cameras, on every `__getitem__`, just to decode a single frame. Random-access sampling re-reads full clip files repeatedly. This is a performance/scalability defect (not a correctness bug): full-file reads per sample make data loading a bottleneck and can OOM on large clips. The author's own inline TODO already flags it.

## Where
- `Model/data_parsing/nvidia_physical_ai/camera.py:105-137` — the per-camera loop in `load_camera_frame`.
  - `:125`: `video_data = io.BytesIO(video_path.read_bytes())  # TODO: major bottleneck for training`.
  - `:126`: `reader = SeekVideoReader(video_data=video_data)`.
  - `:128-131`: decodes a single frame then closes the reader.
  - The loop runs once per camera in `CAMERA_NAMES` (`:21-29`), i.e. 7 full-file reads per `__getitem__`.

## What's wrong / why it matters
`video_path.read_bytes()` loads the whole camera clip into memory, wraps it in a `BytesIO`, constructs a `SeekVideoReader`, decodes exactly one frame, and discards the reader — repeated for all 7 cameras every sample. Costs:
- I/O amplification: to read one frame the loader transfers the entire clip, × 7 cameras × every sample in an epoch. Random-access ordering re-reads the same clip in full many times.
- Memory: peak RAM per `__getitem__` scales with clip file size × concurrent decodes; with multiple DataLoader workers and large clips this can OOM.
- The reader is rebuilt and torn down every call, so no decode state or file handle is reused.

This directly makes GPU training input-bound during NVIDIA data ingest.

## Repro / trigger condition
Any use of the NVIDIA dataset path that calls `load_camera_frame` per sample. Not on the default synthetic/smoke path; does not affect correctness (decoded frames are correct). Manifests as slow data loading and high RAM when ingesting NVIDIA clips, worsening with clip size, camera count (7), and worker count.

## Proposed fix
1. Frame-indexed seeking without full-file reads: open the mp4 by path and seek to `frame_idx` rather than materializing the whole file in a `BytesIO` (pass a path/lazy file-like to `SeekVideoReader` if supported).
2. Cache decoded readers per clip: keep a per-clip (or per-clip-per-camera) reader/handle cache keyed by `video_path` so consecutive samples from the same clip reuse an open reader.
3. Offline pre-extraction (the TODO's suggestion): decode/sample the needed frames in a separate data-processing step and store them, aligning with the repo's existing shard-packing pipeline.

## Severity / priority
Performance/scalability, not correctness. Medium — does not block or corrupt training, but a known throughput bottleneck and OOM risk for real NVIDIA-dataset runs, already acknowledged by the in-code TODO. Latent on default (non-NVIDIA) paths.

_Verified against current code (found by an adversarial audit); `camera.py:125` matches verbatim including the author's TODO._

FLagbusted:
Fixes #116

## Root cause
`io.BytesIO(video_path.read_bytes())` in `load_camera_frame` eagerly reads
the entire mp4 into memory before decoding a single frame — 7x per
`__getitem__`. Traced `SeekVideoReader` directly (physical_ai_av SDK): it
wraps PyAV's `av.open()`, which already does lazy, seek-based decoding —
`container.seek()` jumps straight to the keyframe before the target frame
(confirmed via `_build_keyframe_index` and `decode_images_from_frame_indices`
source). The decoder was never the bottleneck.

## Fix
Plain buffered file handle (`open(video_path, "rb")`) instead of
pre-loading bytes into BytesIO — same seek+read interface, stays lazy.

## Verification
As @riita10069 noted, the two properties that matter here:
- **Absence of full-file reads**, not a timing assertion (which would be
  flaky in CI): `Path.read_bytes` spy shows 0 calls with the fix vs 4
  calls (one per `__getitem__`) on the original code.
- **Content correctness** (seeking lands on the *right* frame, not just
  *a* frame): decoded pixel values match the encoded synthetic value at
  the target frame index exactly, and two different timestamps decode
  two visibly different frames.

Re-verified from a clean checkout immediately before opening this PR —
confirmed the test fails against unfixed `main` (4 reads) and passes
against the fix (0 reads), so this isn't testing against a moved target.

Real-world motivation: I'm about to run this path on a single 24GB GPU —
7 full-file reads per `__getitem__` is the difference between completing
and OOMing, not a latent nicety.

7/7 tests pass, ruff clean.


riita10069:
Wow,  Great Contribution, thanks

riita10069 (approved):

riita10069:
Verified and merging. The root-cause analysis and the read_bytes-spy test approach here were excellent — thank you. Two additions were pushed to your branch as maintainer commits: main added `load_front_clip` (same eager `BytesIO(read_bytes())` pattern) after this PR was opened, so the same one-line fix and matching regression tests were applied there. Full test suite (613 passed) verified on GPU before merge. Fixes #116.

PR #129, opened 2026-07-14
fix(data_ingest): NVIDIA calibration never arrives — the run silently falls back to pseudo geometry

gcordova10:
## The bug

`data_ingest` reads the cameras and the calibration differently, in the same function:

```python
# workflows.py:186,189 — cameras and egomotion
with ds.open_file(cf, maybe_stream=True) as f:            # correct

# workflows.py:196,197 — calibration
intr = ds.get_clip_feature(clip_id, "camera_intrinsics")  # no maybe_stream
extr = ds.get_clip_feature(clip_id, "sensor_extrinsics")  # no maybe_stream
```

`PhysicalAIAVDatasetInterface.get_clip_feature` defaults to **`maybe_stream: bool = False`**
(`physical_ai_av/dataset.py:151`). Under the `local_dir` layout `data_ingest` uses, that cache lookup
misses and the SDK raises:

> `filename='calibration/camera_intrinsics/camera_intrinsics.chunk_0000.parquet' not found in cache;
> set maybe_stream=True to enable streaming from HF Hub.`

The `except` on `workflows.py:207` demotes that to `WARN: no calibration for clip {id}` and the
ingest completes.

## What happens next

No `.pkl` is written → `NvidiaAVDataset.projection_spec` returns `None` → the run continues on
**`geometry_type='pseudo'`** — `PseudoProjection`, which describes itself as *"NOT real geometry ...
a learnable spatial prior"*: a single `[3,4]` matrix expanded across all seven cameras and
initialised from noise.

**So the model trains on a learned warp while every reader downstream assumes real f-theta — and the
only trace is one WARN line in a log of thousands.**

**NVIDIA is the only dataset in the pipeline that ships real calibration, and it has never received
it. The f-theta path added in #77/#107 has never been exercised on real data.**

## Reproduction

Real ingest of the first 10 clips (RTX 3090, 14 GB downloaded): **every clip logged the WARN and the
calibration directory was empty.** With the fix, the same ingest writes `intrinsics.pkl` (3,665 B)
and `extrinsics.pkl` (3,544 B), and `projection_spec` returns `type='ftheta'` with **7 per-camera
matrices**.

## The fix

1. **Pass `maybe_stream=True`**, matching the two reads seven lines above.
2. **Raise if the NVIDIA ingest ends with no calibration at all.** For NVIDIA a missing calibration
   is a **failure, not an absence**: silently degrading to a learned prior is exactly what the
   explicit pseudo path exists to prevent — per the comment at `workflows.py:167`, *"never a silent
   real-geometry claim"*.

There is a **second door to the same failure**: the NVIDIA dataset is gated on HF. If the Flyte
`hf-token` secret has not accepted NVIDIA's terms, the ingest fails → same WARN → same pseudo
geometry. The `raise` closes both.

## Scope

One file, +26 / −2. Nothing outside the NVIDIA branch of `data_ingest` changes. The pseudo fallback
stays exactly as it is for datasets that genuinely have no calibration (L2D).

## Tests

`ruff` clean · `mypy` clean (134 files) · `pytest Model/tests`: **415 passed, 6 skipped**.

### DIFF
--- a/Platform/pipelines/workflows.py
+++ b/Platform/pipelines/workflows.py
@@ -193,8 +193,18 @@ def data_ingest(
                     import pickle
                     ds.download_clip_features(
                         clip_id, features=["camera_intrinsics", "sensor_extrinsics"])
-                    intr = ds.get_clip_feature(clip_id, "camera_intrinsics")
-                    extr = ds.get_clip_feature(clip_id, "sensor_extrinsics")
+                    # maybe_stream=True, exactly as the camera/egomotion reads above.
+                    # get_clip_feature defaults to maybe_stream=False, which resolves
+                    # the chunk through the SDK's local cache; under the local_dir
+                    # layout data_ingest uses, that lookup misses and the SDK raises
+                    # "not found in cache; set maybe_stream=True". The except below
+                    # then swallows it into a WARN, projection_spec returns None, and
+                    # the run silently continues on geometry_type='pseudo' — a learned
+                    # prior — while every reader downstream assumes real f-theta.
+                    intr = ds.get_clip_feature(
+                        clip_id, "camera_intrinsics", maybe_stream=True)
+                    extr = ds.get_clip_feature(
+                        clip_id, "sensor_extrinsics", maybe_stream=True)
                     calib_dir = out / "calibration"
                     calib_dir.mkdir(parents=True, exist_ok=True)
                     with open(calib_dir / "intrinsics.pkl", "wb") as f:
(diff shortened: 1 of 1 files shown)


riita10069:
The current behavior was intentional. L2D does not provide camera intrinsics, so pseudo geometry was being used intentionally.
so for now We should prefer to focus on KIT Scenes. I will close #129.

gcordova10:
@riita10069 — totally agree on L2D, and that path really isn't touched here — this PR only
changes the NVIDIA branch of `data_ingest`, so happy to close it if we're focusing on KIT.

The one bit I'd love to carry over, whichever dataset we land on, is the guard: an empty
calibration directory raising instead of quietly falling back to pseudo geometry with a single
WARN line. KIT's calibration comes through the same `data_ingest`, so it isn't NVIDIA-specific —
and it's really your own point, that we can't measure what quietly lies to us. A run happily
training on a learned prior while everything downstream assumes real geometry is such a painful
one to catch later. I'm glad to fold just that guard into the KIT work if it's useful.

gcordova10:
@riita10069 — closing this. With #144 removing the NVIDIA package the
fix has no object left: it only touches the NVIDIA branch of `data_ingest`, and its guard is
literally `if dataset == Dataset.NVIDIA_PHYSICAL_AI`.

The one thing I would still like to carry over is not the code but the principle behind it: an empty
calibration directory should be a failure, not an absence. What made this worth reporting was not
the missing intrinsics, it was that the run *continued* on `geometry_type='pseudo'` and every reader
downstream assumed real geometry — a silent invalidation rather than a crash. Whatever calibration
KITScenes ships, that distinction is worth keeping.

Issue #121, opened 2026-07-12
Parallelize the full pipeline for 100+ episode training (data-prep + training/eval scaling)


riita10069:
## Motivation

The 3-branch model (reactive + reasoning + WM, PR #119) trains correctly, but on a
10-episode / 1037-sample L2D shard the reasoning/WM branches do **not** beat the
reactive-only baseline (held-out 20%: imitation ADE 1.764 m vs full 3-branch 2.218 m;
in-sample: 1.299 vs 1.407). This is the correct behavior at this data scale, not a
regression: the reasoning coupling stays a *learned no-op* (`reasoning_coupling.alpha ≈
0.004` after 60 epochs, from its zero init) because 1037 near-duplicate frames lack the
scenario diversity for teacher-distilled reasoning to carry generalizing trajectory
signal. The reasoning latent is a re-encoding of inputs the planner already consumes
(visual_history / 64-step egomotion history), so it is redundant here.

Validating the reasoning hypothesis requires **100+ episodes** (yaak-ai/L2D has 100,000).
Scaling ~1k → 100k+ samples exposes bottlenecks across the whole pipeline. Today every
stage runs as a **single Flyte pod** with intra-pod process pools; the real lever is
fanning out across pods/GPUs. This issue tracks that scaling work. Related: #113 (eval
data-leak, fixed in #119), #116 (NVIDIA loader OOM), #117 (double-decode).

## Prerequisite (blocks all data-prep sharding)

- [ ] **Global, episode-anchored `sample_id`.** Today `sample_id = f"s{si:08d}"` is a
  positional index into the in-process `_samples` list (`Model/data_parsing/l2d/dataset.py:196-227`,
  `Model/data_processing/reasoning_label_generation/parallel_label.py:86`,
  `Platform/pipelines/workflows.py:402`). Splitting episodes across pods renumbers `si`,
  silently breaking the label↔pack JOIN and colliding every S3 cache key. Replace with a
  hash/compose of `(episode_index, frame_index)` (both already on the sample,
  `dataset.py:319-320`). Unblocks map-task sharding of labeling + packing and makes the
  reasoning-label cache reusable under any sharding.

## Data-prep (ingest / label / pack)

- [ ] **`data_ingest` ephemeral-storage overflow.** `ephemeral_storage=400/450Gi`
  (`workflows.py:129-130`) vs 0.5–7 TB raw video at 100+ episodes; `shutil.copytree`
[...]

PR #130, opened 2026-07-14
feat(reasoning): unpooled BEV as a spatial context source — the architectural prerequisite from #121

gcordova10:
## Why

In #121 you traced the reasoning coupling to a **learned no-op** (`reasoning_coupling.alpha ≈ 0.004`
after 60 epochs, from its zero init) and gave the reason: the head consumes `visual_history` +
`ego_context`, which is a **strict subset** of what the planner consumes (`visual_history` + `ego` +
**pooled** BEV). By the data-processing inequality the reasoning latent cannot carry trajectory
information the planner lacks, so `alpha → 0` is the optimizer being **correct**, not a bug.

And you named the one principled fix:

> *"Feed the reasoning head the UNPOOLED spatial BEV that the bezier planner discards via
> `bev_features.mean(dim=(2,3))`. Attending the full `[B,C,H,W]` BEV gives the reasoning latent
> spatial hazard information the mean-pooled planner context cannot reconstruct."*

**This PR implements that.** Without it, the reasoning branch we are about to merge is — by your own
analysis — a branch the model provably ignores.

## What this adds

**`BevContextTokenizer`** — turns the unpooled BEV `[B, C, H, W]` into spatial context tokens:
adaptive-average-pool to a coarse grid (default 16×16), project each cell to `hidden_dim`, add a
learned 2-D positional embedding.

Attending the BEV cell-per-token is not viable — the default grid is 450×300, i.e. **135k cells**. The
coarse grid keeps *where* a hazard is (which the mean-pool destroys) while keeping the token count
sane for a 1 Hz branch. **The grid size is the ablation knob.**

## Design — it fits your head without touching the decoder

The BEV enters `HorizonReasoningHead` as a **third context source**, alongside the `route_context_dim`
/ `map_context_dim` slots **you already left open**. It is the only *spatial* source, so it
contributes a **sequence** of tokens (`cat`) rather than a single vector token (`stack`) — but the
decoder interface is unchanged: only the context set grows.

`fused_features` is already computed at the call site in `ReactiveE2E`, so **no reordering is needed**.

### Two independent switches, deliberately not one

- **`reasoning_bev_input`** — the head **sees** the spatial BEV.
- **`reasoning_bev_detach`** — whether the reasoning loss also **reshapes** the shared trunk through
  that path.

This repo has **twice** found that routing auxiliary gradient into the shared backbone **hurts at low
data** (the JEPA loss floor → `detach_backbone`; the non-zero-init `visual_history_proj`). Flipping
both at once would make a result unattributable, so **see-but-don't-reshape is the default** and the
gradient path is opt-in — as you asked for in #121.

## Also: `train_il` never passed `reasoning_kwargs`

`train_il` builds `AutoE2E` without threading `reasoning_kwargs` (`workflows.py:793-798`), so
`HorizonReasoningHead`'s optional context slots — `route`, `map`, and now the BEV — **cannot be turned
on by any training run in the repo**. The feature exists and is unreachable.

This PR threads three flat scalars (`reasoning_bev_input`, `reasoning_bev_grid`,
`reasoning_bev_detach`) rather than a dict, because those are the knobs a Flyte run is launched with
from the CLI and a dict parameter cannot be set there.

When the BEV is on, the run also **prints each shard's geometry**. The whole argument for the unpooled
BEV is that it carries *where* a hazard is; on a dataset without calibration the grid is a learned
warp and a cell maps to nowhere, so a null result there would be uninterpretable. Better a loud line
in the log than a silently worthless run.

## Cost — measured, and it is the capacity control

| Arm | Params | Peak, `bs=1` (RTX 3090, fp32, WM off) |
|---|---|---|
| imitation only | 91.35 M | 9.97 GB |
| reasoning, current inputs | 93.91 M | 9.98 GB |
| **reasoning + unpooled BEV** | **94.04 M** | **9.98 GB** |

**The BEV costs +0.132 M params (+0.14 %) and under 10 MB.** That is deliberate: **if the BEV arm beats
the no-BEV arm, it cannot be capacity — there is no capacity to attribute. It can only be the
information.** Fixed *ex ante*, before any run.

Grid as the ablation knob: 8×8 → +0.083 M · 16×16 → +0.132 M · 32×32 → +0.328 M.

## Strictly additive

Built without `bev_context_dim`, the head is **byte-identical to today** (`test_head_without_bev_
ignores_the_bev_argument`). Default off.

## Scope — not in this PR

- The episode-count sweep and the `alpha` curve. That needs the data scaling #121 tracks, and I'd
  rather report a curve than promise a number.
- The `sample_uid` prerequisite — that's yours, and I won't duplicate it.

## Tests

Four files, +256 / −3. `ruff` clean · `mypy` clean (135 files) · `pytest Model/tests`: **422 passed**.

The tests pin the hypothesis itself: **two BEVs with the same per-channel mean but a different spatial
layout must produce different latents.** If the head cannot tell them apart, the extra tokens are
decoration and the ablation is pointless. Plus the two switches (see-only leaks no gradient into the
trunk; the gradient path does), and the byte-identical-without-the-BEV check.

### DIFF
--- a/Model/model_components/reactive_e2e.py
+++ b/Model/model_components/reactive_e2e.py
@@ -77,10 +77,24 @@ class ReactiveE2E(nn.Module):
         self.enable_reasoning = enable_reasoning
         self.reasoning_mode = reasoning_mode if enable_reasoning else "none"
         self.ReasoningHead = None
+        # TWO INDEPENDENT SWITCHES, deliberately not one:
+        #   bev_context_dim  -> the head SEES the unpooled BEV (#121's fix).
+        #   reasoning_bev_detach -> whether the reasoning loss also RESHAPES the
+        #       shared BEV/backbone through that path.
+        # This repo has twice found that routing an auxiliary gradient into the
+        # shared trunk HURTS at low data (the JEPA floor -> detach_backbone; the
+        # non-zero-init visual_history_proj). Flipping both at once would make a
+        # result unattributable, so the gradient path is opt-in and OFF by default.
+        self.reasoning_bev_detach = True
         if enable_reasoning:
             rkw = dict(reasoning_kwargs or {})
             rkw.setdefault("visual_history_dim", visual_history_dim)
             rkw.setdefault("ego_context_dim", egomotion_dim)
+            self.reasoning_bev_detach = bool(rkw.pop("bev_detach", True))
+            # The BEV the head attends is the FUSED map+image BEV, so its channel
+            # count is embed_dim. Derive it rather than making the caller repeat it.
+            if rkw.get("bev_context_dim") is True:
+                rkw["bev_context_dim"] = embed_dim
             self.ReasoningHead = HorizonReasoningHead(**rkw)
 
         # Trajectory decoder — swappable via planner_mode (gru, flow_matching).
@@ -155,9 +169,20 @@ class ReactiveE2E(nn.Module):
         reasoning_latent = None
         reasoning_horizon_tokens = None
         if self.ReasoningHead is not None:
+            # fused_features is already computed above, so the BEV is sitting at
+            # the call site: no reordering needed. Detached by default -> the head
+            # SEES the spatial BEV without the reasoning loss reshaping the shared
+            # trunk (see the two-switch note in __init__).
+            bev_context = None
+            if self.ReasoningHead.bev_tokenizer is not None:
+                bev_context = (
+                    fused_features.detach() if self.reasoning_bev_detach
+                    else fused_features
+                )
             reasoning_pred = self.ReasoningHead(
                 visual_ctx, ego_ctx,
                 route_context=route_context, map_context=map_context,
+                bev_context=bev_context,
             )
             reasoning_latent = reasoning_pred.reasoning_latent
             reasoning_horizon_tokens = reasoning_pred.horizon_tokens
--- a/Model/model_components/reasoning/horizon_reasoning_head.py
+++ b/Model/model_components/reasoning/horizon_reasoning_head.py
@@ -69,6 +69,54 @@ class AttentionPool(nn.Module):
         return pooled.squeeze(1)                   # [B, D]
 
 
+class BevContextTokenizer(nn.Module):
+    """Turn the unpooled BEV ``[B, C, H, W]`` into spatial context tokens.
+
+    Motivation (the #121 root cause): the head's inputs (``visual_history`` +
+    ``ego_context``) are a strict subset of the planner's, so by the
+    data-processing inequality its latent cannot carry trajectory information the
+    planner lacks — and the coupling correctly learns ``alpha ~ 0``. The one
+    signal the planner *discards* is the BEV's spatial structure: the bezier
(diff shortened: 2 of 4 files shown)


riita10069:
I am opposed to this kind of architectural change.

My position is that, if the Planner already uses spatial BEV information sufficiently, then providing the same BEV to Reasoning is not necessary merely to increase the information available to the model.
Reasoning may still provide value by producing a semantic and temporal re-representation of the same observations under a separate teacher signal. It does not necessarily need privileged observations that are unavailable to the Planner.


More fundamentally, the Reasoning branch is intended for long-tail situations. Its value should emerge with data scale and with training examples that contain sufficiently complex and rare scenarios. At present, we have not trained at a meaningful scale, and we have not trained adequately on long-tail situations.

I will ask @m-zain-khawaja for his judgment on

gcordova10:
@riita10069 — good call to be cautious here, so I went and tested it before pushing further,
and it's worth sharing because the result surprised me too.

This PR is really your idea from #121 — give the reasoning head the un-pooled BEV the bezier
planner throws away — so I built it and ran it as a clean ablation. Same model, same NVIDIA
clips, same labels, same F-Theta geometry, same seed. The only thing that changes between the
two arms is whether the head sees the spatial BEV.

| | alpha | intervention delta | ADE | FDE |
|---|---|---|---|---|
| reasoning, current inputs | -0.000334 | 1.7e-06 | 3.528 | 9.797 |
| reasoning + spatial BEV (this PR) | +0.000735 | 2.8e-06 | 4.321 | 11.381 |

The coupling doesn't open. Both alphas are basically zero — the 0.004 you measured on L2D is
5x larger than either — and manipulating the reasoning latent moves the trajectory by 2.8e-06.

The part I found interesting: I wanted to be sure a flat result meant "the BEV doesn't help"
and not just "the head ignored the BEV", so I checked whether the head actually uses it. I fed
it two BEVs with identical per-channel means but different spatial layouts. The pooled BEV the
planner sees (`bev_features.mean(dim=(2,3))`) is mathematically identical for both — only
something reading the *where* can tell them apart.

```
relative change in reasoning_latent under a pure spatial permutation:
  trained head (this PR)          0.0425
  same architecture, untrained    0.0013     (31.9x)
```

So the head does read the spatial BEV and learned to — and the planner still ignores its
latent. Which is the interesting bit for both of us: on this data, information wasn't the
binding constraint. We removed exactly the redundancy your #121 argument points at, and the
gate didn't move. Something else is holding it shut, and I genuinely don't know what yet —
could be the labels, the zero-init coupling at this scale, or a training-objective mismatch.

So I'm not asking to merge this as an improvement — it's the ablation that produced the number,
not a fix, and the hypothesis didn't survive. Happy to close it, or keep it around if it's
useful for reproducing the result; it's opt-in and off by default either way. Your call.

Your other point — that the branch needs data scale and long-tail examples — I think is the
right next question, and it's one we can actually test rather than assume. I've been poking at
the labels for that (the teacher only sees the front camera right now, which may be capping
what the head can learn), and I'll write it up separately so this thread stays about the BEV.
Curious what you make of the permutation result in the meantime.

PR #131, opened 2026-07-14
feat(reasoning): produce the teacher embeddings the alignment loss already expects (#98)

gcordova10:
## The gap

`HorizonReasoningLoss` accepts `teacher_embedding_targets [B, 5, D]` for its alignment term
(`horizon_reasoning_loss.py:224,234`), `schema.py` declares the field on `ReasoningTargetBatch`, and
`HorizonReasoningHead` produces `student_reasoning_embedding [B, 5, D]` to match it.

**But nothing in the pipeline ever produces the targets.** Neither `record_to_target_tensors` nor
`target_batch_from_loader` ever fills them, so `teacher_embedding_targets` is always `None` and **the
alignment term never contributes.**

The loss was written against a producer that does not exist. This adds it.

## Where it goes, and why there

`record_to_target_tensors` is the **last point in the pipeline that still holds the per-horizon
`evidence` text the teacher wrote**. The loader flattens records to class indices immediately after,
and the text is gone.

So encoding it there is the **smallest wiring that makes the alignment loss reachable at all** — no new
pipeline stage, no new shard member.

## Design

- `teacher_encoder` is an **injected object**, not a hard dependency: `HashingTextEncoder`
  (deterministic, no model download, safe in CI) or `SentenceTransformerEncoder`.
- Per horizon: encode `evidence` → `[5, D]`, plus a `valid [5]` mask. An abstained horizon (or one
  with no evidence) is masked out — **it never contributes a target**, matching the R9 abstain rule.
- **Opt-in.** With no encoder, `record_to_target_tensors` output is **byte-identical to today** and the
  alignment term stays inert exactly as it is now.

## Fails loudly rather than degrading

A **mixed batch** — some samples embedded, some not — **raises** rather than dropping the term for the
whole batch. Dropping it silently would leave training running, the alignment loss doing nothing, and
nobody any the wiser. That is the failure mode this branch exists to remove, not to add.

## Open question, not a decision — your loss, your call

The embedding is computed **at load time** from the evidence text. That is cheap and deterministic with
`HashingTextEncoder`; a `SentenceTransformer` would repeat the work every epoch.

**The alternative is computing it once at pack time and storing it in the shard** — bigger shards,
computed once.

Happy to move it: the seam is one argument either way. I went with load-time because it needs no
change to the packer or the shard format, but you own this loss and I'd rather you choose.

## Scope

Three files, +375 / −2. Nothing changes for a run that doesn't pass an encoder.

## Tests

`ruff` clean · `mypy` clean (136 files) · `pytest Model/tests`: **428 passed**.

Covers: byte-identical without an encoder; the emitted `[5, D]` + `valid [5]` shapes; that
`collate_reasoning_targets` produces the `[B, 5, D]` the loss consumes; the abstain/no-evidence mask;
and that a mixed batch raises instead of silently dropping the term.

### DIFF
--- a/Model/data_processing/reasoning_label_generation/targets.py
+++ b/Model/data_processing/reasoning_label_generation/targets.py
@@ -60,14 +60,28 @@ def _group_attr(horizon: ReasoningHorizonLabel, group: str) -> Any:
 
 
 def record_to_target_tensors(
-    record: ReasoningLabelRecord, taxonomy: ReasoningTaxonomy = DEFAULT_TAXONOMY
+    record: ReasoningLabelRecord,
+    taxonomy: ReasoningTaxonomy = DEFAULT_TAXONOMY,
+    teacher_encoder: Any = None,
 ) -> Dict[str, torch.Tensor]:
     """Tensorize one record into per-sample targets (no batch dim).
 
     Returns a flat dict:
         ``target__<group>``: ``[5, C]`` float (multi) or ``[5]`` long (single),
         ``confidence``:      ``[5]`` float,
-        ``source_weight``:   ``[5]`` float.
+        ``source_weight``:   ``[5]`` float,
+    and, when ``teacher_encoder`` is given:
+        ``teacher_embedding``:       ``[5, D]`` float,
+        ``teacher_embedding_valid``: ``[5]`` float (1 = usable, 0 = masked).
+
+    ``teacher_encoder`` closes a gap: ``HorizonReasoningLoss`` already accepts
+    ``teacher_embedding_targets [B, 5, D]`` for its alignment term, but **nothing
+    in the pipeline produced them**, so the term was inert. The record carries the
+    per-horizon ``evidence`` text the teacher wrote, and this is the one place that
+    still has it — the loader flattens records to class indices right after, and the
+    text is gone. Encoding it here is therefore the smallest wiring that makes the
+    alignment loss reachable at all. It is OPT-IN: with no encoder, the output is
+    byte-identical to today.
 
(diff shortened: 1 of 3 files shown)


riita10069:
My understanding from #98 was that we had agreed not to implement this component.
So I will close it.

gcordova10:
@riita10069 — I think we might be talking about two different things here, and I want to check
it against the thread rather than trust my memory, because I could be the one misremembering.

What we set aside in #98 was Intisar's idea of feeding `student_reasoning_embedding` **to the
planner** as a control path — on 8 July you said it'd be "mostly redundant as a planner input"
and that you "would not make that part of the initial implementation", and I agreed the
planner-facing interface stays `horizon_tokens` / `reasoning_latent`.

But the alignment loss itself is your §5:

```
L_alignment = mean_h (1 - cosine_similarity(student[h], teacher[h]))
student_reasoning_embedding: [B,5,D_teacher]
```

and in that same 8 July comment: *"So I would start with `L_align(student_reasoning_embedding,
teacher_reasoning_embedding)` as an auxiliary training loss only."* You even built it —
`HorizonReasoningLoss._alignment(teacher_embedding_targets: [B,5,D])` with a passing test, and
`lambda_alignment` defaulting to `0.0` because nothing feeds it yet. So the training-only
auxiliary head you designed is sitting there untrained.

This PR is just the missing producer for it — three files, all under
`reasoning_label_generation/`, and "planner" doesn't appear anywhere in the diff. It's not the
control path we set aside; it's the targets your alignment loss is already waiting for.

That said, you have more context on the #98 discussion than I do — if you're thinking of
something I'm missing, point me at it and I'll close it myself, no problem.

PR #132, opened 2026-07-14
fix(model): reject removed geometry kwargs instead of silently dropping calibration

gcordova10:
## Problem

`Model/data_parsing/kit_scenes/forward_pass_test.py` does not run against the current
model. Both of its modes crash:

```
test_bev_infer_mode   ValueError: not enough values to unpack (expected 3, got 1)
test_bev_train_mode   ValueError: not enough values to unpack (expected 3, got 1)
```

It is not wrong — it follows `Model/README.md`, which still documents the pre-#77/#107
signature: `camera_params=` and a 3-tuple return. Neither exists any more.

The crash has an obvious fix — delete the third return value — and that fix is the real
problem. `forward` takes `**kwargs` and hands them to `ReactiveE2E.forward`, which hands
them to the planner; all three absorb them. So `camera_params=<calib>` is dropped without
a word and the model falls back to `geometry_type="pseudo"`, which `projection.py:268`
calls "NOT real geometry, a learnable spatial prior (shape-testing / ablation only)".

Measured on main, same weights and inputs:

| call | trajectory[0, :3] |
| --- | --- |
| no geometry | `0.21758756, 0.15876013, 0.19457541` |
| `camera_params=<calib>` | `0.21758756, 0.15876013, 0.19457541` |
| `projection=PinholeProjection(<same calib>)` | `0.21374298, 0.15579164, 0.19097323` |

Bit for bit identical: the calibration never entered the computation. The control shows
real geometry does move the output, so the route is broken, not the calibration. Fix the
crash and the test passes, prints a plausible trajectory, and throws KITScenes'
calibration away. The `ValueError` is the only thing warning us, and it is the first
thing anyone will delete.

No training run is affected: `train_il` passes `projection=` + `geometry_type=`
correctly, and `e2e_pipeline_smoke.py:135` asserts `"camera_params" not in batch`.

## Fix

- **`Model/model_components/auto_e2e.py`** — `forward` raises `TypeError` on the removed
  geometry kwargs (`camera_params`, `calib`, `intrinsics`, ...), naming the replacement.
  An error rather than a warning, because the failure mode is a silently wrong run and
  nobody reads a warning in a five-hour training log. `**kwargs` stays open for the
  planner; only the geometry names are rejected.
- **`Model/README.md`** — the real signature and the projection ABI.
- **`kit_scenes/forward_pass_test.py`** — ported to `projection=` + `geometry_type=` and
  to the 1-or-2 return contract. It also unpacked a `planner_loss` that `forward` has
  never returned.
- **`Model/tests/test_auto_e2e.py`** — 9 tests pinning the guard.

Not here: wiring KITScenes end to end (it still needs a `projection_spec`, and its map
tiles need Lanelet2). This only makes its forward test run, and makes a dropped
calibration impossible to miss.

## Verification

ruff + mypy clean (91 files), `pytest Model/tests` 299 passed / 4 skipped.

### DIFF
--- a/Model/README.md
+++ b/Model/README.md
@@ -14,22 +14,46 @@
 
 **AutoE2E outputs a prediction of:**
 - Future driving trajectory (modelled as future acceleration and curvature values over a 6.4s future horizon at 10Hz sampling rate)
-- `ego_hidden` — 256-dim final GRU hidden state from `TrajectoryPlanner` summarising the planner's intent over the prediction horizon. Conditions `FutureState`; replaces the legacy compressed visual feature vector / rolling visual history buffer.
 
 **During training, and for purposes of model introspection, AutoE2E also predicts:**
-- Future visual features at 1.6s intervals for a 6.4s future horizon (what does the future feature representation of the scene look like, this is used for a feature reconstruction loss similar to JEPA)
+- Future visual features at 1.6s intervals for a 6.4s future horizon (what does the future feature representation of the scene look like, this is used for a feature reconstruction loss similar to JEPA). Opt-in via `enable_world_model=True`.
 
 **Forward signature:**
 ```python
-trajectory, ego_hidden, future_visual_features = model(
-    visual_tiles,        # (B, V, 3, H, W) — 7 cameras
-    map_tile,            # (B, 3, H_map, W_map) — BEV nav-map image
+trajectory = model(
+    camera_tiles,        # (B, V, 3, H, W) — 7 cameras
+    map_input,           # (B, 3, H_map, W_map) — BEV nav-map image
     visual_history,      # (B, 896) — frame-to-frame visual memory
     egomotion_history,   # (B, 256)
-    camera_params=None,  # (B, V, 3, 4) optional, used by BEV fusion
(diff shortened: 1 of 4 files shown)


riita10069:
Thanks for identifying this.

My understanding is that #132 only affects callers using the legacy camera_params=... API. The current training path already passes projection=... and geometry_type=..., so this issue does not affect the training configuration we are using for the v1 measurable baseline. is it correct?

gcordova10:
Exactly — the current training path is unaffected, which is why I flagged it in the PR as not a
live bug. The one caller that still uses the legacy `camera_params=` API is
`kit_scenes/forward_pass_test.py` — on `main` now, since the merge: it crashes on the 3-value
unpack today, and the obvious fix (dropping the third return value) would make it pass while
quietly falling back to pseudo geometry. So this is really just to make that fail loudly instead
of silently, before the KIT work leans on that test. Happy to keep it low-priority for the v1
baseline. (I've rebased the PR onto the merged `main`, so it's conflict-free again.)

m-zain-khawaja (approved):
Approved - thanks @gcordova10

PR #133, opened 2026-07-14
fix(tests): run the NVIDIA FTheta calibration tests that CI has always skipped

gcordova10:
## Problem

`TestBuildFThetaFromCalibration` (4 tests) has never run in CI: `requirements.txt` has
no scipy, so `importorskip("scipy")` skips it every time. `build_ftheta_projection` is
green by default, not by test.

Adding scipy alone makes them error instead of skip: importing
`nvidia_physical_ai.calibration` pulls `camera.py`, which imports the NVIDIA SDK at
module scope, and the SDK is not in CI.

## Fix

- **`camera.py`, `egomotion.py`** — import the SDK inside the functions that use it. It
  is needed to decode video, not to import the package; `calibration.py` is pure numpy.
- **`test_projection.py`** — `importorskip("scipy", minversion="1.16")`. The tests use
  `RigidTransform`, new in 1.16; on 1.15 the guard passed and the test died on
  `AttributeError` instead of skipping.
- **`requirements.txt`** — `scipy==1.16.3`, `pandas==2.3.3`.

No test added; `build_ftheta_projection` is unchanged and correct.

## Verification

All four pass on Python 3.12. They were not failing — they were not running.
`pytest Model/tests` 290 passed locally (scipy < 1.16, so they skip cleanly). ruff +
mypy clean.

scipy 1.16 needs Python >= 3.11, so `make setup` now needs 3.11 (CI runs 3.12). If you
would rather not move that floor, the test can stub the pose — `calibration.py:67` only
calls `.as_matrix()` on it.

### DIFF
--- a/Model/data_parsing/nvidia_physical_ai/camera.py
+++ b/Model/data_parsing/nvidia_physical_ai/camera.py
@@ -15,7 +15,6 @@ from pathlib import Path
 import numpy as np
 import pandas as pd
 import torch
-from physical_ai_av.video import SeekVideoReader
 
 # Camera directories present in the NVIDIA PhysicalAI-Autonomous-Vehicles dataset.
 CAMERA_NAMES: list[str] = [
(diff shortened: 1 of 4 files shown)


gcordova10:
@riita10069 — closing this. It makes `TestBuildFThetaFromCalibration` run in CI, and those
tests import `build_ftheta_projection` from the package #144 removes
(`test_projection.py:378,410,423`), so the target goes away with it.

One consequence outliving the PR, worth flagging so it does not disappear with the close:
`requirements.txt` still declares no `scipy`, and after the NVIDIA removal `scipy` is still imported
by KITScenes — `Model/data_parsing/kit_scenes/egomotion.py:22`
(`from scipy.spatial.transform import Rotation`), reached through
`workflows.py:1313 -> kit_scenes/__init__ -> dataset.py -> egomotion.py`. It works today only
because the environments happen to have it. That is an undeclared runtime dependency on the path
that survives, so it stops being an NVIDIA-testing detail and becomes a KIT one — leaving it written
here so whoever does the #144 sweep can fold it in.

PR #137, opened 2026-07-18
Benchmark-5080RTX

CodeName-Detective:
The benchmark was run on an NVIDIA RTX 5080 machine using the script from the `main` branch at the time the fork was created.

Please confirm whether the benchmark script on `main` has since been updated to include the World Model. If so, I will rerun the benchmark using the updated script and revise the results accordingly.


CodeName-Detective:
I did't follow the requested standards, hence I resubmitted as another pull request.

PR #139, opened 2026-07-18
bench: add 5080RTX results

CodeName-Detective:
The benchmark was run on an NVIDIA RTX 5080 machine using the script from the `main` branch at the time the fork was created.

Please confirm whether the benchmark script on `main` has since been updated to include the World Model. If so, I will rerun the benchmark using the updated script and revise the results accordingly.


riita10069:
Thank you for contributing the RTX 5080 re-run — since the existing 5080 section in BENCHMARKS.md was measured on PyTorch 2.11.0 / commit `9015914` and yours is PyTorch 2.7.1 / `ead2171`, it was kept as a separate section labeled as a re-run rather than replacing the original. Running the reasoning variants (`pooled_latent` / `horizon_cross_attention`) as well is much appreciated.

One maintainer commit was pushed to your branch before merging: per-GPU tables live in `Model/speed_benchmark/BENCHMARKS.md` (see the note in `Model/speed_benchmark/README.md`), not the root README, so your table was moved there — rendered from your results JSON, including the Reasoning column. The JSON itself was merged untouched.

A follow-up request, if you still have access to this hardware: the World Model branch (`enable_world_model=True`) now exists in `AutoE2E` but the benchmark script does not measure it yet. If you are interested, extending `speed_benchmark.py` with a world-model variant (mirroring how the reasoning variants were added) and re-running on your GPUs would make these tables cover all three branches. Happy to review that as a separate PR.

Issue #146, opened 2026-07-21
[Feature]: KITScenes Camera Input Views


m-zain-khawaja:
### Problem / motivation

Optimization of the camera views for KITScenes

### Proposed solution

- KITScenes dataset has six surround view cameras and a long-range front-facing camera
- We can drop the ring-front camera view since it is almost identical to front-center and additionally add the long-range front-facing HD camera to be able to view the scene at a greater distance

<img width="997" height="378" alt="Image" src="https://github.com/user-attachments/assets/cb82cc0f-5315-4ee3-adbb-add0282caef8" />

m-zain-khawaja:
@gcordova10 - could you please take this one

gcordova10:
Taking this one — PR is #150.

Measuring the cameras first turned up that the long-range camera is `camera_base_front_center`,
which we already feed. So dropping `ring_front` is right, but on its own it does not extend range:
every view is resized before it reaches the model, and the long-range camera arrives no sharper than
a ring camera.

What does extend it is the tele view you described first — a 30 deg centre crop of that camera at
native resolution, **2.9x the angular resolution of any full view**. Seven views, as you proposed.
Calibration, derivation and tests are in the PR.

One practical note: `num_views` comes from the packed shard manifests, so **the shards need repacking** 
for the new view set before the next training run.

PR #150, opened 2026-07-21
feat(kit_scenes): drop the redundant ring-front view and add a tele crop of the long-range camera (#146)

gcordova10:
## Summary

- Drops `camera_ring_front` from the KITScenes view set: it is redundant with the long-range front
  camera, measurably so.
- Adds `tele_center`, a 30 deg centre crop of that same long-range camera taken at native
  resolution, so the model actually gains reach.
- Still 7 views. No pipeline, model or `image_size` changes.

## Why

**Which camera is the long-range one.** The calibration published in the SDK's [`notebooks/04_calibration_and_multimodal.ipynb`](https://github.com/KIT-MRT/kitscenes/blob/7765cdec5490894266070ab46e23724b58b3da42/notebooks/04_calibration_and_multimodal.ipynb), at the revision this repo pins (`KITSCENES_SDK_REVISION`), gives, with HFOV = 2·atan(W / 2·focal):

| camera | W x H | MPix | HFOV |
|---|---|---|---|
| `camera_base_front_center` | 5856 x 3104 | 18.2 | 88.2 deg |
| `camera_ring_*` (6) | 3504 x 2272 | 8.0 | 87.1 deg |
| `camera_base_front_*_rect` (2) | 2272 x 3488 | 7.9 | 63.3 deg |

Those three groups line up with the sensor suite in the dataset paper (arXiv:2606.02956): one
long-range camera at 88.4 deg, six surround cameras at 87.1 deg, and a tilted stereo pair at
63.3 deg. So the long-range imager is `camera_base_front_center` — which the model already feeds as
the first entry of `CAMERA_NAMES`.

**Why `camera_ring_front` goes.** It points the same way as the long-range camera and covers
essentially the same field of view at 2.3x fewer pixels. The two are redundant, so one of them is
paying compute for nothing.

**Why dropping it is not enough on its own.** Reach is set by the pixels per degree that reach the
model, not by the sensor's pixel count, and every view is resized to `image_size` (256 by default):

```
camera_base_front_center   5856 px over 88.2 deg  ->  256 px  ->  2.90 px/deg
camera_ring_front          3504 px over 87.2 deg  ->  256 px  ->  2.94 px/deg
```

At 256 px the long-range camera is no sharper than a ring camera — the resize discards exactly the
resolution that would have provided the extra range. Feeding it whole cannot extend how far the
model sees.

**What does.** Cropping before the resize keeps it. A 30 deg centre crop is 1619 px of the 5856 px
frame and lands at 256 px as **8.54 px/deg — 2.9x any full view**. Cropping does not change focal
length, so the crop is treated as a narrower camera sharing the source pose; only the principal
point moves. The crop is square, matching the square resize, and centred on the principal point so
the optical axis stays in the middle of the frame.

The alternative would be raising `image_size` instead. Matching 8.54 px/deg on the full frame needs
753 px of input rather than 256 — 8.6x the pixels per view, across all seven. The crop reaches the
same angular resolution at today's cost.

The result keeps peripheral context (the long-range camera whole, plus 5 ring cameras) and adds
distance (the same camera cropped), at the same 7 views as before.

The two are not co-located, so this does drop a viewpoint — but nothing in the model uses stereo or
disparity, so no path loses information. The crop's field of view is one constant
(`TELE_VIEW_HFOV_DEG`) if you want it narrower or wider.

## Validation

- `Model/tests/test_kitscenes_tele_view.py` (new, 8 tests): the crop spans the requested FOV, is
  centred on the principal point, fits inside the frame, and rejects a FOV that does not. Two tests
  pin the reason the view exists: the crop carries >2.5x the full frame's pixels per degree, and the
  full long-range frame is no sharper than a ring camera once resized.
- `ruff` clean. The KITScenes and view-fusion suites pass: 86 passed, 6 skipped.
- `num_views` is **derived from the packed shard manifests**, not hardcoded, so the change
  propagates on its own — but **existing shards must be repacked** for the new view set.

### DIFF
--- a/Model/data_parsing/kit_scenes/README.md
+++ b/Model/data_parsing/kit_scenes/README.md
@@ -35,15 +35,29 @@ Without Lanelet2, map tiles fall back to zero tensors and the dataset still func
 
 ## Model inputs produced
 
-- `visual_tiles` `(6, 3, H, W)` — 6 camera frames
+- `visual_tiles` `(7, 3, H, W)` — 6 camera frames plus the derived tele view
 - `map_tile` `(3, H, W)` — BEV map tile (rasterization of Lanelet2 HD map)
 - `egomotion_history` `(256,)` — fixed model ABI: 64 past timesteps × 4 signals at 10 Hz
 - `visual_history` `(896,)` — zero-initialised placeholder; populated during sequential inference
 - `trajectory_target` `(128,)` — fixed model ABI: 64 future timesteps × 2 signals
-- `camera_params` `(6, 3, 4)` — projection matrices `P = K_scaled @ T_ref_to_cam` for the 6 camera views, computed from KITScenes calibration and scaled to match the backbone's resize/crop transform. 
+- `camera_params` `(7, 3, 4)` — projection matrices `P = K_scaled @ T_ref_to_cam` for the 7 views, computed from KITScenes calibration and scaled to match the backbone's resize/crop transform. 
 
 
-The 6 camera views are the long-range front camera (`camera_base_front_center`) plus 5 surround ring cameras (`CAMERA_NAMES` in `camera.py`). `camera_ring_front` is dropped as it duplicates the long-range camera's forward coverage (#146), and the stereo front pair stays out: at 63.3 deg HFOV it trades field of view for baseline.
+The 7 views are 6 physical cameras plus one derived view (`VIEW_NAMES` in `camera.py`):
+
+| view | source | HFOV | px/deg at 256 |
+|---|---|---|---|
+| `camera_base_front_center` | long-range camera, 5856x3104 | 88.2 deg | 2.90 |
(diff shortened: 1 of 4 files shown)


m-zain-khawaja (approved):
Approved - thanks @gcordova10

riita10069:
This PR is currently broken and should be reverted.

`KitScenesDataset` still defaults to `CAMERA_NAMES`, which now contains only six physical cameras. As a result, the normal packing path never generates `tele_center` and produces six views instead of seven.

Changing the default to `VIEW_NAMES` is not enough, because `tele_center` is a derived view and the Dataset currently treats every entry as a physical SDK camera. That would cause scenes to be skipped.

The cache contract versions were also not updated, so the pipeline may reuse incompatible old shards or mix partitions with different view counts. In this state, the intended data cannot be generated reliably and the training pipeline failed.

Issue #152, opened 2026-07-22
[Feature]: Lanelet2 to BEV Map rendering update


m-zain-khawaja:
### Problem / motivation

We need to encode route information in our map input to the network when we are relying upon a Lanelet2 map. 

### Rendering
- We should render the Lanelet2 map in a 256px x 256px square grid of resolution 0.5m
- We can take the ground truth waypoints from the vehicles motion and create a one-hot (black/white) image encoding the route data of the same size and resolution as the map render
- Both of these inputs can be fed into the network to train
- The route should be encoded up to the edge of the map view
- If it is possible, we can try snapping the waypoints to the centre-lines taking car at lane changes, where for example, if the driven path doesn't snap neatly then we do some interpolation.

<img width="416" height="179" alt="Image" src="https://github.com/user-attachments/assets/6f468c73-1f9a-477c-8e69-5bcb57d713b0" />

### Inference
- Prior to an actual autonomous driving run, a user would have to annotate the route they want the vehicle to take offline and we would use that to create the one-hot encoding of the route

m-zain-khawaja:
@bharatwrrr - could you please take this on

bharatwrrr:
@m-zain-khawaja, I think #161 closes this issue as well.

PR #161, opened 2026-07-25
Add route-conditioned navigation inputs for KITScenes (#149)

### LINKED ISSUE #149: [Feature]: GPS to BEV map rendering updated
### Problem / motivation

- Our model backbones expect square 1:1 aspect ratio images of size 256px by 256px however, our GPS to BEV map rendering [script renders](https://github.com/autowarefoundation/auto_e2e/tree/main/Model/data_parsing/map_rendering) a map on a 640px by 360px grid, which will cause warping and scaling artefacts when fed directly into the vision encoder backbones.
- Furthermore, we should explicitly encode the context of the scene and separate it from the navigation goal

### Proposed solution

- We should please adjust the map rendering to be on a 1:1 square tile of size 256 x 256 pixels covering a grid area of 256m front, back, left and right, equivalent to 1m per pixel.
- The vehicle should near the bottom centre of the map, 1/3 up the height at a bottom offset of 85px.
- We should rasterize the open-street map data with all key contexts including roads, traffic lights, direction of traffic, vegetation, building structure and we can leave out the building names or shop names
- We should create a separate black-white image where the white pixels denote a line encoding the future path the vehicle should follow to reach its goal, e.g. going straight, turning right etc. We can render the path that fits on the map grid.
- We need to take care to align the direction of the map such that it matches the direction of the vehicle, instead of always pointing North. This way the map route and context can be directly used by the network as a prior without the network needing to transform the map internally in its features. This can be done by aligning the map with the yaw angle of the route such that the route always points up at the start
- We should consider how we feed these two images into the map encoder, perhaps the simplest way is to add an additional channel to the RGB rasterized image - please investigate

Image

### For inference, we should download the vector for a region and do on-board real-time rasterization:

**Client-Side Real-Time Vector Rendering:**

Instead of generating image files on a server, the server delivers raw geospatial coordinates (points, lines, polygons) as lightweight vector packets. The user’s browser or mobile device then draws the map in real time using the device's graphics card.How it works: The client requests .pbf or .mvt vector tiles, and client-side JavaScript libraries style and render them dynamically.The Backend: You can host an instance of TileServer GL or look into the community-driven [OpenMapTiles](https://openmaptiles.org/) stack.The Frontend: Use web frameworks like [MapLibre GL JS](https://maplibre.org/) or Mapbox GL JS to manipulate colors, hide layers, or switch map languages instantaneously on the fly.Pros: Smooth infinite zooming, tiny server storage footprint, and instant theme changes (like switching to dark mode).

### Route Planning:

Route planning is out of scope - we should assume that a navigation route to reach a goal is pre-computed by the user or by a routing service such as Google Maps, etc. We simply need to ensure that we can render the route on the black/white image.

### Alternatives considered

_No response_

riita10069:
## Summary

This PR implements the route-conditioned navigation input proposed in #149.

It introduces provider-independent navigation contracts, offline KITScenes route generation, a native BEV rasterizer, explicit model inputs, runtime route handling, and route-aware evaluation.

## Architecture

Training:

KITScenes scene trace
→ Lanelet2 map matching
→ selected lanelet sequence
→ NavigationRoute
→ C++ BEV rasterizer
→ lossless KITScenes v3 shard members

Inference:

user destination
→ local Valhalla service
→ NavigationRoute
→ 2 Hz C++ BEV rasterization
→ 10 Hz model input using the latest valid raster

The model input contract is:

```text
map_context : float32 [B, 14, 256, 256]
route_mask  : float32 [B,  2, 256, 256]
map_valid   : bool [B]
route_valid : bool [B]
```

The route channels are:

- channel 0: selected lanelet corridor
- channel 1: destination marker

The model processes them as follows:

```text
map_context ─ map_valid gate ─┐
                              ├─ channel concat
route_mask ─ route_valid gate ┘
    → shared SwinV2 NavigationEncoder
    → zero-initialized residual fusion with camera BEV
    → Reactive planner
```

Route information is intentionally restricted to the Reactive branch. The Reasoning branch continues to receive only visual history and ego context.

## Actual KITScenes v3 route inputs

The following image shows raw 256 × 256 route masks from ten distinct KITScenes v3 scenes. Magenta visualizes the selected corridor and yellow visualizes the destination channel. The colors are only for inspection; the model receives two binary channels.

KITScenes v3 route masks

## Leakage controls

Ground-truth future waypoints are never rasterized directly. Complete scene traces are map-matched to Lanelet2 geometry using position, heading, traffic rules, and route continuity. The rasterizer receives only map geometry, selected lanelet IDs, destination, and ego pose.

## Validation

- KITScenes v3 publication is ready with 404 shards and 42,667 packed samples.
- Actual published shards were decoded and inspected.
- Route-conditioned training produced non-zero gradients for both the route input and NavigationEncoder.
- Rasterizer parity, coordinate, loader, model-boundary, runtime, Flyte, and evaluation tests are included.
- DCO passes; GitHub CI is running for the documentation-asset commit.
- Controlled paired evaluation is still in progress, so this PR does not yet claim an accuracy improvement.

Closes #149.


bharatwrrr:
Hi @riita10069. I really like the switch to a 14-channel map image and the BEV rasterizer you added with it. Is it safe to say this also closes #152? Thanks!

riita10069:
Training result update

We completed the controlled comparison using the same KITScenes v3 artifacts, frozen validation split, training seed, and training configuration, with route conditioning as the only intended model-input difference.

| Condition | Best epoch | ADE (m) | FDE (m) |
|---|---:|---:|---:|
| No-route baseline | 7 | 3.7823 | 10.9128 |
| Route-conditioned | 9 | 3.7805 | 10.5128 |

The route-conditioned model changed ADE by only -0.0018 m (-0.05%) and improved FDE by 0.4000 m (3.7%). This is not a material improvement in aggregate ADE, and the FDE gain is modest. We therefore do not claim that the current training objective makes strong use of the route input.

The route input and NavigationEncoder both received non-zero gradients, so the input path itself is working. Extending training has not improved validation so far:

| Epoch | ADE (m) | FDE (m) |
|---:|---:|---:|
| 9 | 3.7805 | 10.5128 |
| 10 | 3.8863 | 11.0351 |
| 11 | 4.0903 | 11.8569 |

After Epoch 11, `bad_epochs=2`; Epoch 12 is running and the existing early-stopping policy will terminate the run if it does not improve.

This PR still establishes the provider-independent navigation contract, deterministic KITScenes route generation, semantic BEV rasterization, Flyte data path, Reactive-only model input, and route-aware evaluation needed for further research. I plan to follow this with a stronger training objective that combines:

- a camera-only BEV segmentation auxiliary loss before map fusion;
- mean-normalized long-horizon trajectory weighting with decay 0.99;
- deterministic junction- and maneuver-aware training resampling;
- a differentiable selected-route consistency loss.

Each component will remain independently configurable so the combined result can be ablated. Given that the current limited result is documented and the infrastructure is functioning as designed, I will merge this PR and continue the model-quality work in a follow-up.

m-zain-khawaja:
I think this is a fascinating result:

- we should also explore how we do the map fusion (e.g. residual vs attention based)
- also, how did the network train with just the reactive branch vs the other branches added on - do you have those metrics? 
- I think to simplify our training, we can start with training just the reactive model and improve it as much as possible and then consider adding in the other branches one by one.

PR #176, opened 2026-07-31
Align KITScenes planning loss with rollout evaluation and composite checkpoint selection

riita10069:
## Summary

This PR aligns KITScenes training and checkpoint selection with the trajectory behavior evaluated after control rollout.

- Retains the normalized action SmoothL1 objective and adds differentiable XY path/final-position losses through the shared semi-implicit unicycle rollout.
- Adds target-relative comfort, drivable-area, and selected-route constraints without using future trajectory geometry as route supervision.
- Adds a reconstruction audit that measures how well target acceleration and curvature reproduce logged future motion before rollout supervision is enabled.
- Reports both natural and `split_group_uid` scene-balanced validation metrics.
- Selects checkpoints with a weighted trajectory, comfort, map-safety, and navigation score while independently retaining the best scene-balanced ADE/FDE checkpoint.
- Stops only when neither composite score nor trajectory quality improves, with a default patience of five epochs.
- Persists selector policy, metric semantics, coverage, resume state, and both best-checkpoint roles to MLflow and checkpoint metadata.
- Keeps the KITScenes model input at the six physical camera views and stores loss-only route/drivable supervision during Flyte preprocessing.

The main motivation is that action-only loss and ADE/FDE-only checkpoint selection do not directly reward the rollout, comfort, map safety, and selected-route behavior that matter during planning.

## Matched training result

The comparison below uses the same frozen KITScenes validation split: 40 scenes and 3,820 samples.

| Metric | Previous best | Latest best |
|---|---:|---:|
| Composite score | 0.4453 | 0.5527 |
| Trajectory utility | 0.5242 | 0.5614 |
| Comfort utility | 1.0000 | 0.9997 |
| Map-safety utility | 0.0000 | 0.2632 |
| Navigation utility | 0.1663 | 0.4128 |
| Scene-balanced logged-XY ADE@3s | 1.3397 m | 1.0744 m |
| Scene-balanced logged-XY FDE@6.4s | 13.6700 m | 11.8959 m |
| Control-rollout FDE@3s | 3.0737 m | 2.6124 m |
| Mean off-road excess | 0.3244 | 0.2799 |
| Mean route gap | 0.4087 | 0.3547 |
| Mean wrong-branch excess | 0.7184 | 0.7126 |
| Mean destination error | 10.8387 m | 9.3519 m |
| Post-training ADE/FDE@3s | 4.4517 / 12.2064 m | 3.9041 / 10.9361 m |

The latest run improved the composite score by 24.1%, scene-balanced ADE@3s by 19.8%, control-rollout FDE@3s by 15.0%, and the post-training ADE/FDE pair by 12.3% / 10.4%. Comfort remained effectively unchanged.

MLflow runs:

- Previous: `e643849d97a44d52b3648a0ab3275d25`, best epoch 2
- Latest: `8e238504ab354e7f8c7828acb8f51b1f`, best epoch 18

The completed matched run used selector policy v2, where the long-horizon scene-balanced FDE component was still measured at 6.4 seconds. This PR upgrades the canonical selector and MLflow contract to policy v3, where both scene-balanced logged-XY ADE and FDE are measured at 3 seconds. A new Full Run is therefore required for the final selector-v3 benchmark.

## Validation

- GitHub Actions `lint-and-test`: passed
- Ruff: passed
- mypy: passed
- Unit tests: 895 passed, 14 skipped, 6 deselected
- EC2 GPU suite before enabling Flyte-only modules: 790 passed, 6 skipped, 6 deselected
- Reconstruction audit and rollout forward/backward benchmarks are included in the committed artifacts and design document.


m-zain-khawaja:
@riita10069 - Let's please use X,Y for ADE/FDE, I think it more accurately captures how well the car is driving vs ground truth and is easy to understand.

riita10069:
@m-zain-khawaja  thanks zain, I will merge it, if you want to revert it, I will do.
basically I want to keep autoe2e simplest, lowest source code.


## List of Changes at the Requirements Level

### 1. Align the training objective with 3-second vehicle rollout performance

* The control values output by the model, such as acceleration and steering, must be passed into a **differentiable planar bicycle model** and converted into a vehicle trajectory.
* The rollout duration must be **3.0 seconds**, with a time step of **0.1 seconds** in principle.
* The training objective must cover not only raw control values, but also the following errors after rollout:

  * XY position error
  * Heading error
  * Curvature error
  * Temporal smoothness of controls
  * Comfort based on acceleration and jerk
  * Deviation from the selected route
  * Deviation from the drivable area
  * Proximity and collision risk with other vehicles
* The rollout-aligned loss must be integrated into the existing trajectory imitation loss, and **the evaluation target used during training must be aligned with the target used for checkpoint selection**.
* Heading comparisons must account for the ±π angular wraparound boundary.
* Items for which supervision is unavailable must not be incorrectly treated as zero loss; an availability mask must be used to exclude them from calculation.
* For route, drivable-area, and collision losses, the number of applicable samples and the supervision coverage must be recorded.

Primary implementation: [rollout_aligned_loss.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/training/losses/rollout_aligned_loss.py), [control_rollout.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/training/losses/control_rollout.py)

### 2. Add supervision data for the selected route and drivable area

* The selected navigation route must be explicitly represented as supervision data for each sample.
* The following information must be generated and stored as `route_supervision.npz`:

  * Distance field to the selected route corridor
  * Distance field to the drivable area
  * Sine and cosine of the route heading
  * A mask indicating cells where the route heading is valid
  * Destination coordinates in the ego frame
  * Destination visibility
  * Availability of drivable-area supervision
* This supervision must be used **only for loss calculation** and must not become a dependency of the model input at inference time.
* `map_semantic.npz`, `route_mask.npz`, `route_supervision.npz`, and `navigation_meta.json` must be treated as a complete set, and partially present sets must not be accepted.
* Data whose navigation geometry ID does not match the model-side geometry contract must be rejected.
* For data without navigation supervision, values must not merely be zero-filled; `available=false` must also be set explicitly.

Primary implementation: [supervision.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/navigation/supervision.py), [artifacts.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/navigation/artifacts.py)

### 3. Update the packed dataset data contract

* The packed-shard contract must be updated from **schema v5 to v8**.
* The sample navigation contract must be updated from **v1 to v2**.
* The shard must include `route_supervision.npz` and `split_group_uid` for scene-level aggregation.
* Artifacts created under old contracts or stale artifacts must not be silently accepted as data under the new contract.
* Existing v5 packed data is not fully backward-compatible as-is, and in principle must be **repacked and regenerated**.
* The data loader must validate schema v8 completeness, shapes, finite values, and geometry IDs.

Primary implementation: [contract_versions.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/data_processing/contract_versions.py), [pre_extracted.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/data_parsing/pre_extracted.py)

### 4. Increase training exposure for turns, intersections, and similar cases

* Rather than training directly on a natural distribution dominated by straight driving, samples where route choice is important must be deterministically oversampled.
* When the selected route is valid and the maneuver is one of the following, the sample must in principle be exposed **four times**:

  * left
  * right
  * u-turn
  * merge
  * exit
* Intersection samples must in principle be exposed **two times**.
* The maximum exposure count per sample must be **four**.
* This repeat policy must be applied only to the training split and must not be applied to the validation split.
* Repetition must occur before shuffling and image decoding, so that the effective training distribution is reproducible.
* The system must make it possible to audit not only the number of unique samples, but also the effective exposure count after repetition, maneuver-specific counts, junction-specific counts, and the exposure digest.

This is not merely a model-loss change. It is a requirement-level change to **the training distribution itself** in order to strengthen route-decision learning.

### 5. Fix validation metrics to 3-second logged XY

* Checkpoint evaluation must use a **3.0-second horizon** as the formal selection basis.
* The evaluation coordinate frame and target semantics must be fixed to **logged XY**.
* XY trajectory errors at 1, 2, and 3 seconds must be recorded.
* The following metrics must be calculated from the same rollout:

  * XY ADE / FDE
  * Heading error
  * Curvature error
  * Comfort
  * Route violation
  * Drivable-area violation
  * Collision rate
* Aggregation must not use a simple sample mean. It must first average within each scene and then average across scenes using a **scene-balanced mean**.
* Long scenes containing more samples must not be allowed to dominate checkpoint selection excessively.

Primary implementation: [rollout_validation.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/evaluation/rollout_validation.py)

### 6. Use a composite score for checkpoint selection

* A single metric such as “minimum validation loss” or “minimum ADE” must not be used alone to determine the best checkpoint.
* Checkpoints must be ranked using a composite score that integrates trajectory, heading, curvature, comfort, route, drivable-area, and collision metrics.
* Each metric must be normalized against a reference distribution and combined using configured weights.
* The selector policy must be versioned as `rollout-ckpt-v2`.
* A fingerprint must be generated from the selector configuration so that selectors with the same name but different configurations can be distinguished.
* The score, rank, and component metrics for each checkpoint must be stored as a leaderboard artifact.
* Monotonicity analysis must make it possible to verify whether performance improves in the expected direction across checkpoint scans and training-data increments.

Primary implementation: [checkpoint_selection.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/evaluation/checkpoint_selection.py)

### 7. Do not select checkpoints based on differences in supervision coverage

* A checkpoint must not be treated as a valid candidate when coverage for required selector metrics is zero or missing.
* When coverage for route, drivable-area, collision, or similar metrics differs between checkpoints, those checkpoints must not be compared as-is.
* It must be verified that the same evaluation population and coverage are secured across all checkpoints.
* The system must prevent selection of a checkpoint that appears to have a good route metric only because it has less route supervision and therefore receives a smaller penalty.
* Coverage values must also be retained in selector artifacts and model metadata.

This is an important part of the PR: **not only the metric values, but also the population over which those metrics were computed, are formalized as part of the contract**.

### 8. Link early stopping to the selector score

* Early-stopping decisions must be aligned with the new rollout checkpoint-selector score.
* Selection state, including the best checkpoint, best score, epoch / step, and patience, must be persisted.
* If the training process is interrupted, the best-selection state must be restorable from saved state.
* After resume, the best score and early-stopping counter must not be reset in a way that produces a different selection result.
* The final best checkpoint must not change depending on whether a crash and resume occurred.

The PR description refers to this as “crash-safe early stopping and resume-state restoration.”

### 9. Enforce the selector contract during model registration

* A checkpoint registered in the model registry must satisfy the following:

  * The selector policy version matches the expected version
  * The horizon is 3.0 seconds
  * The target frame is `logged_xy`
  * The aggregation method is `scene_balanced_mean`
  * The metric namespace matches the expected value
  * A selector configuration fingerprint exists
* A checkpoint selected under an out-of-contract evaluation method must not be registered as the official best model.
* The following must be stored in registration metadata so that the selection result can later be reconstructed:

  * Best checkpoint path
  * Checkpoint digest
  * Epoch / step
  * Selector score
  * Leaderboard rank
  * Selector policy / configuration fingerprint
  * Dataset / split / training provenance
* The system must make it possible to trace not only **which checkpoint was registered**, but also **why that checkpoint was selected**.

Primary implementation: [training_checkpoint.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Platform/pipelines/training_checkpoint.py)

### 10. Add a reconstruction audit for target controls

* Target controls in packed shards must be integrated with the bicycle model and compared against logged future XY reconstructed from GPS / pose data.
* The audit must be executable without decoding camera or map tensors, using only the following minimum required artifacts:

  * `ego.npy`
  * `gps.npy`
  * `pose.npy`
  * `meta.json`
* Duplicate, missing, or inconsistent sample UIDs and split-group UIDs must be detected.
* ADE / FDE must be calculated for both the 3-second horizon and the full 6.4-second horizon.
* The current default thresholds are:

  * p95 of 3-second FDE: **1.0 m or less**
  * p95 of 6.4-second FDE: **2.0 m or less**
* Alignment between the heading derived from controls and the logged-path tangent must also be audited.
* Results must be produced at the sample level, scene level, each timestep, and for the worst-performing scenes.
* Digests of the sample-UID set and scene-UID set must be recorded so that use of the same evaluation population can be verified.
* When thresholds pass, the automatic recommendation must be `go`; when they fail, it must be `review_required`. However, the final decision must remain `pending_review` so that human judgment is preserved.

Primary implementation: [reconstruction_audit.py](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/evaluation/reconstruction_audit.py)

### 11. Fix the identity of the training and validation data

* The train / validation split must be fixed using a versioned manifest rather than an implicit split defined only in code.
* `kitscenes_train_dev_v3.json` must be used so that split groups, dataset contracts, source identities, and related information are traceable.
* Validation scenes must not enter the training repeat policy.
* Digests of the split, dataset, and artifacts must be retained in training metadata and audit metadata.
* The same validation-group set must be used when comparing checkpoints.

Primary implementation: [kitscenes_train_dev_v3.json](https://github.com/autowarefoundation/auto_e2e/blob/26f2c06404a849df34bb55978f177294b525a0bd/Model/training/splits/kitscenes_train_dev_v3.json)

### 12. Keep the camera-input contract at six physical views

* The model input must continue to use **six physical camera views**.
* An additional synthesized view must not be introduced into the parser contract as a new seventh camera.
* The number and ordering of camera-parser views must be fixed through regression tests.

This is less a new feature and more a constraint to prevent the current inference-facing camera contract from being broken by the other changes in this PR.

### 13. Update the pipeline, documentation, and tests for the new contract

* The training pipeline must execute rollout validation, checkpoint ranking, artifact upload, and metadata registration as one integrated workflow.
* Checkpoint selection and reconstruction auditing must be executable as independent pipeline tasks.
* Selector score, policy, coverage, and best-checkpoint information must be recorded in MLflow.
* The release-readiness check must verify compliance with the new selector and audit contracts.
* Regression tests must be added for the following:

  * Bicycle rollout
  * Rollout loss
  * Route consistency
  * Navigation artifact encoding / decoding
  * Stale artifact rejection
  * Checkpoint ranking
  * Coverage validation
  * Resume-state restoration
  * Model-registration metadata
  * Reconstruction audit
* Procedures for checkpoint selection, reproducible metadata, and reconstruction auditing must be documented for operators.

PR #182, opened 2026-08-02
feat(evaluation): report the reasoning-branch intervention delta as an eval metric

gcordova10:
## Problem

`reasoning_intervention_delta` (added in #109) measures whether the planner actually uses the reasoning
branch: bypass the branch, re-run the same batch, and report how far the trajectory moves. This is the
gate `g` discussed in #123. But nothing calls it during evaluation, so **no checkpoint reports it** —
today `g` is a manual, one-off measurement. #123 asked for the opposite: a number computed *per
checkpoint* that answers "is the reasoning used?" on its own, and rises by itself if the coupling ever
opens with data scale.

One caveat shaped the wiring: the function's forward did not pass the navigation inputs (#161) or a
fixed `initial_noise`. So a plain call would measure the delta off the real operating point — and, for
a stochastic planner, with different noise in the two runs, which swamps the signal.

## Fix

- **Extend the harness.** `reasoning_intervention_delta` / `horizon_intervention_delta` take
  `**forward_kwargs`, threaded identically into both runs — a fixed `initial_noise` (so both runs share
  it) and the navigation inputs (so the delta is at the real operating point). Additive and
  backward-compatible.

- **Wire it into eval.** `_evaluate_open_loop` gains a `report_intervention` flag (default off).
  `evaluate_il_policy` turns it on; the training-time validation leaves it off, so **training cost is
  unchanged**. When on and the model has a reasoning head, it computes the delta on up to 50 batches
  (reusing that batch's stable `initial_noise`), averages, and returns `reasoning_intervention_delta`;
  `_run_evaluation` logs `eval/reasoning_intervention_delta` next to ADE/FDE.

- **Guarded.** With no reasoning head the metric is skipped, so this is a byte-identical no-op for
  non-reasoning checkpoints.

## Verification

- `ruff` clean; `mypy` clean on `Model/evaluation/faithfulness.py`.
- `Model/tests/test_faithfulness_forward_kwargs.py` (new, 3 tests): a plain call is noise-dominated; a
  fixed `initial_noise` recovers only the intervention; the navigation inputs are threaded.
  `test_reasoning_faithfulness.py` stays green — **11 passed** together.
- `Model/tests/test_workflow_training_lifecycle.py` (3 new tests): the metric is reported when opted in,
  absent without a reasoning head, and absent by default. These reuse the file's existing
  `flytekit`-gated eval harness, so they run in CI alongside the other workflow tests.

## Coordination

As I flagged in #123, this touches `_evaluate_open_loop`, which #176 (`feat/rollout-aligned-checkpoint-selection`) also restructures. I offered there to base it on that branch or open against `main` and rebase once it lands — opening against `main` here so it is not blocked, and I will rebase onto #176 when it merges so the metric reports inside the composite selector. Happy to coordinate on the exact wiring point.

Refs #123 (the gate this computes) and #109 (the harness it calls).

### DIFF
--- a/Model/evaluation/faithfulness.py
+++ b/Model/evaluation/faithfulness.py
@@ -50,6 +50,7 @@ def reasoning_intervention_delta(
     projection: Optional[Any] = None,
     geometry_type: Optional[str] = None,
     image_transform: Optional[Any] = None,
+    **forward_kwargs: Any,
 ) -> dict[str, float]:
     """Mean trajectory L2 between the reasoning-coupled and bypassed runs.
 
@@ -57,6 +58,12 @@ def reasoning_intervention_delta(
         model: an ``AutoE2E`` built with ``enable_reasoning=True``.
         camera_tiles / map_input / visual_history / egomotion_history: one batch.
         projection / geometry_type / image_transform: current geometry ABI.
+        forward_kwargs: extra forward inputs threaded identically into both runs —
+            navigation (``route_mask`` / ``map_valid`` / ``route_valid``) so the
+            delta is read at the checkpoint's real operating point, and a fixed
+            ``initial_noise`` so a stochastic planner uses the SAME noise in the
+            coupled and bypassed runs (otherwise the delta is noise, not the
+            intervention).
 
     Returns:
         ``{"trajectory_l2": float}`` — 0.0 while the coupling gate is untrained.
@@ -76,7 +83,7 @@ def reasoning_intervention_delta(
     model.eval()
     restore_buffer = _snapshot_buffer(model)
     fwd = dict(projection=projection, geometry_type=geometry_type,
-               image_transform=image_transform, mode="infer")
+               image_transform=image_transform, mode="infer", **forward_kwargs)
 
     try:
         with torch.no_grad():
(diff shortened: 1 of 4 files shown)


riita10069:
https://github.com/autowarefoundation/auto_e2e/issues/168#issuecomment-5143205649

gcordova10:
Rebased onto main now that #176 has landed, and MERGEABLE again. Both conflicts were additive rather than competing: each side added a keyword to _evaluate_open_loop, and #176 replaced the ADE/FDE computation with the 3-second window plus per-horizon accumulation, which this metric sits alongside rather than duplicates. ruff and mypy clean, 920 passed.

The description said I would rebase "so the metric reports inside the composite selector". The rebase makes the two coexist, but it does not put the delta into the selector: the composite score is computed in the per-epoch validation, where this metric is deliberately off. Wiring it there would add a per-epoch cost the current design avoids, so it reports next to ADE/FDE in _run_evaluation instead. Happy to wire it either way if you prefer.

On the scoping point: this does not add a reasoning layer or change what is trained — it is skipped when the model has no reasoning head, and the delta is capped at 50 samples, so Reactive-only runs take the path they take today. If you would rather park it until the reasoning layer is picked back up, that is fine by me; the rebase is done either way.

PR #190, opened 2026-08-12
bench: add RTX 4060 8 GB inference speed results

FLagbusted:
nothing else it it, Adds RTX 4060 8 GB inference speed benchmark results only.


riita10069 (approved):

riita10069:
@FLagbusted Thanks

PR #193, opened 2026-08-12
Feed reasoning confidence into the planner coupling

### LINKED ISSUE #110: [Feature]: Make reasoning-band confidence supervised and consumed by the planner (completes #103 safety loop)

ShauryaVM:
Follow-up to #110 / the #103 safety loop.

The reasoning band already emits confidence (and we supervise it with Brier), but the planner coupling ignored it. This scales the zero-init residual by pooled confidence so low confidence means weaker reasoning modulation — i.e. stay closer to the reactive baseline. At init `alpha` is still 0, so any confidence value remains a no-op until the gate opens.

Also wires confidence through `ReactiveE2E` → Bezier / FlowMatching, and adds a small ECE helper for eval.

### How I tested
- `pytest Model/tests/test_reasoning_coupling.py`

Closes #110


riita10069:
Thank you for your contribution.
I would like to know the A/B test results.
I'm not sure if this change is worth merging.

riita10069:
Under the contribution standard proposed in #198, I consider this PR AI slop in its current form. It changes learned model behavior but provides no training comparison, ADE/FDE results, or analysis of model performance. This is not ready for review; please provide the required real-world validation before resubmitting.

ShauryaVM:
A/B numbers for the confidence→planner change.

Trained two mock Combined-style models for 12 steps on the same batch (CPU): **scaled** (this PR) vs **unscaled** (confidence forced to 1, i.e. pre-#110). JSON: `Model/evaluation/results/confidence_coupling_ab.json`.

| Arm | ADE@3s | FDE@3s | loss 0→12 |
|--|--|--|--|
| scaled (this PR) | 2.3927 | 5.2045 | 4.032 → 0.978 |
| unscaled (conf=1) | 2.3928 | 5.2050 | 4.032 → 0.978 |
| Δ (scaled − unscaled) | **−0.00012 m** | | |

Forcing conf=0 vs conf=1 on the scaled model after those steps also moves ADE by ~0. That is the containment property: `alpha` is still ~0, so the planner is byte-identical until the gate actually learns. Once the gate is open, the existing unit test shows low confidence shrinks the residual (high-conf displacement > low-conf).

I am not claiming a KITScenes ADE win. The A/B says this change does **not** regress open-loop ADE while the coupling is still a no-op, which is the bar for landing the safety-loop wiring.

PR #195, opened 2026-08-12
Action-conditioned JEPA future predictor (optional action_dim)

ShauryaVM:
Small wedge toward an action-sensitive world model.

`FutureFeatureMapPredictor` / `WorldActionModel` now take an optional `action_dim`. When set, actions are projected into the same seed space as the history and added as a residual (Delta-JEPA style). The action projection is zero-initialised, so at init the forecast matches the history-only path for any action — same containment idea as the reasoning coupling gate.

Default construction is unchanged (`action_dim=None`). Opt in via `world_model_kwargs={"action_dim": 2}` (or whatever control dim you use) and pass `actions=` into `predict_future` / `forward`.

### How I tested
- `pytest Model/tests/test_world_action_model.py` (23 passed), including zero-init no-op and post-open sensitivity checks

### DIFF
--- a/Model/evaluation/action_jepa_ab.py
+++ b/Model/evaluation/action_jepa_ab.py
@@ -1,0 +1,216 @@
+"""A/B: history-only JEPA vs action-conditioned JEPA in the train loop.
+
+Trains two Combined mock models on the same batch:
+  * ``history`` — ``action_dim=None`` (current Combined default)
+  * ``action``  — ``action_dim=128``, AutoE2E teacher-forces ``trajectory_target``
+
+Reports JEPA loss after the same step count, whether ``action_proj`` left zero,
+and matched-vs-shuffled action JEPA on the action-conditioned arm (the wiring
+check: at init the residual is a no-op so the two JEPA values match; after
+training they should split if the projection used the plan).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import torch
+
+
+ACTION_DIM = 128  # 64 timesteps × 2 signals (a, κ)
+
+
+def _batch(device: torch.device, b: int = 2, v: int = 6, hw: int = 256) -> dict[str, torch.Tensor]:
+    return {
+        "visual": torch.randn(b, v, 3, hw, hw, device=device),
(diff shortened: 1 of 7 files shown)


riita10069:
Under the contribution standard proposed in #198, I consider this PR AI slop in its current form. It adds an architecture and API without wiring them into the training loop or reporting any trained JEPA result or model-performance evaluation. This is not ready for review; please provide the required real-world validation before resubmitting.

ShauryaVM:
The action residual is now in the training loop, not just the WAM API.

`AutoE2E.forward` teacher-forces `trajectory_target` into `predict_future` when `world_model_kwargs.action_dim` is set. `train_il` / `wf_train_il` take `action_conditioned_jepa` (default off) which builds `action_dim=128` (flattened 64×(a,κ) plan) and stores it on the checkpoint so eval rebuilds the same architecture.

A/B on a Combined mock batch, 12 AdamW steps, CPU (JSON: `Model/evaluation/results/action_jepa_ab.json`):

| | JEPA | notes |
|--|--|--|
| history-only Combined | 0.0725 | `action_dim=None` |
| action-conditioned | 0.0688 | Δ −0.0037 vs history-only |
| same model, shuffled actions | 0.0688 | Δ vs matched ≈ 0 on random frames |

Containment holds: at init, matched vs shuffled actions are byte-identical (pred L1 = 0, JEPA 0.0982 both). After those 12 steps `action_proj` L2 = 2.40 and swapping actions moves the forecast (pred L1 = 0.010). Gradients reach `action_proj` through `jepa_loss`.

I am not claiming a KITScenes JEPA win — future frames in this batch are random, so shuffled actions barely change reconstruction. The number that matters for landing the wiring is: the residual is a no-op until trained, then the train loop actually uses the plan.

Same CLI on a packed partition:

```
python -m evaluation.action_jepa_ab --shard-dir /path/to/partition --out action_jepa_ab.json
```

How I tested: `pytest Model/tests/test_action_jepa_ab.py Model/tests/test_world_action_model.py` (4 + 23 passed).

PR #200, opened 2026-08-19
fix(model): vectorize egomotion normalization and stop mutating the caller

ChristophBank:
## The bugs

`normalize_egomotion` scaled speed and acceleration by editing the caller's tensor **in place**, in a Python loop over every batch element and timestep.

**Correctness.** The tensor belongs to the DataLoader batch, not to `forward`. Because the scaling isn't idempotent, a second forward pass over the same batch divides speed by 33 again. `Platform/pipelines/overlay_precompute.py`'s multi-seed loop reuses one batch across seeds and hits this exactly: every seed after the first reads an already corrupted speed value.

**Speed.** 1,024 serialized scalar index ops per forward at batch 8. A Python loop is doing what should be one tensor op.

auto_e2e_ego_normalization_mutation_bug

## The fix

Return a new tensor instead of mutating the input, and replace the per-element loop with one broadcast multiply. Output values are numerically identical to the loop (verified in the added test), only *how* the scaling happens changes.

## Where the benchmark numbers come from

I measured on my RTX 5080: `torch.cuda.Event` timing, 20 warmup iterations + 200 timed calls per point, at batch sizes 1/8/32. The loop scales with batch size, the vectorized version doesn't:

pr_egomotion_loop_vs_vectorized

## Testing

`Model/tests/test_egomotion_normalization.py` covers: the input is not mutated, calling it twice on the same *output* is expected to differ (confirming it's the aliasing that was the bug, not the scaling itself), which columns get scaled, batch-size coverage, dtype/device preservation, gradient flow, and the width guard.


ChristophBank:
Closing to redo — commits weren't signed and need to be. Will reopen once that's fixed.

